PR #2, opened 2012-05-17
fix typo so that post sanitized options_ used as intended.  Also practic...

matthewvon:
...e git skills before big stuff.


jonmeredith:
+1 merge

PR #3, opened 2012-05-18
Write value of executing options to log upon database open.

matthewvon:
Diz requested that all options at the time of database open be placed in LOG to allow technical review.  This only logs global options, not the read and write specific options.

### DIFF
--- a/db/db_impl.cc
+++ b/db/db_impl.cc
@@ -140,6 +140,8 @@ DBImpl::DBImpl(const Options& options, const std::string& dbname)
 
   versions_ = new VersionSet(dbname_, &options_, table_cache_,
                              &internal_comparator_);
+
+  options_.Dump(options_.info_log);
 }
 
 DBImpl::~DBImpl() {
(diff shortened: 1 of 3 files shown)


rzezeski:
This worked fine for me but I created a branch to write names where possible instead of just emitting pointers which seems mostly useless.

+1 to merge

https://github.com/basho/leveldb/compare/rz-mv-option-logging-use-names

rzezeski:
Also, as a note, the cacheleak test fails for me intermittently on this branch and master.

matthewvon:
Read and Write options are not available to leveldb at time of database open.  They are available only at time of an actual read or write.  Was not in the mood to create necessary code to ensure options only written the first time encountered ... i.e. I got enough performance concerns this round without creating another.  Maybe next release.  The underlying problem though is solved.  Critical options that were previously NOT passing from app.config to leveldb are now verified to pass.

rzezeski:
I'm not sure I follow.  I was referring to the `comparator` and `filter_policy` options.  Both of them have a human readable string representation available yet you log the pointers.

PR #4, opened 2012-05-19
Add multiple compaction threads with different task priorities.  Also mo...

matthewvon:
...ve file close and unmap operations to background threads.  This update ends almost all stalls (subsequent cache checkin fixes low file handle stalls).  This can slow down throughput due to compactions happening as scheduled (subsequent file sizing addresses throughput).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ build_config.mk
 *.so.*
 *_test
 db_bench
+*~
--- a/db/db_impl.cc
+++ b/db/db_impl.cc
@@ -627,7 +627,7 @@ void DBImpl::MaybeScheduleCompaction() {
   }   // if
 
 
-  if (bg_compaction_scheduled_ && NULL==imm_) {
+  if (bg_compaction_scheduled_ && !push) {
     // Already scheduled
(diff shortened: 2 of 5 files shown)


chardan:
Notes for further work (the nitpick department-- not things you need to change, but things I don't want to get lost):
- Suggest -Wall
  (Note: looks like it compiles mostly cleanly, some signed/unsigned comparison warnings.)
- As per conversation with Matthew, it might be nice in a future date to peek at Boost threads, which
  supply things like scoped_mutex and may make it easier to eventually align with standard C++11 
  threads. Likewise boost::function/C++11 std::function. Out of scope for this release.
- In general, a lot of bare pointers (including function pointers) could probably go away (for
  example, by using smart pointers and functors; this may even slightly improve performance in the latter
  case particularly); I'm not sure that's worth it as long as pthreads are used, however. 
- the code could use some clean up and general TLC; feels slightly inconsistent
- exception handling might well be worth considering in this code as a whole; for instance, operator new and container operations using it
  (like push_back()) can all throw. It may be worth exploring C++11, some operations might be snappier (for instance,
  by avoiding extra copy.

chardan:
+1. Nice work, Matthew!

PR #5, opened 2012-05-20
Clean up caching for better read speed and better size accounting.

matthewvon:
Following changes:
1. reader reader/writer lock objects for use with cache so GET operations can fly through quicker
2. changed LRU accounting from rearranging list on every read (which required write lock) to timestamp based
3. made LRUCache a proper subclass of Cache object so it could be used independently
4. table_cache now uses LRUCache instead of sharded cache because distribution of files across shards was highly uneven for low file counts (i.e. Riak usage) and caused stalls
5. block_cache now includes size of LRUHandle and key in its size accounting

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ build_config.mk
 *.so.*
 *_test
 db_bench
+*~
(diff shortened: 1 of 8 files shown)


chardan:
- It would be nice in a future pass to see if some of the void pointers can be eliminated in favor of something typesafe (I think it will be possible).
- -Wall reveals a few tidbits, it would be nice if it were the normal flag.

chardan:
+1 (please peek at the break vs. flag thing)

PR #6, opened 2012-05-29
Mv badblock recovery

matthewvon:
This change set is to address the issue where a corrupt block can put compaction into an infinite loop.


jonmeredith:
Code review + testing found a couple of issues that are now resolved.
Running a longer test to check compaction working ok, will +1 on completion.

jonmeredith:
+1 merge - a couple of corruptions got stuck in snappy, but not levels fault.

PR #8, opened 2012-05-29
Mv perf counters

matthewvon:
This change creates one memory mapped file with a series of incrementing counters.  External programs can also map this file so as to publish them through other monitoring channels such as snmp.


gburd:
+1, ready to merge.

PR #13, opened 2012-06-20
write throttle

matthewvon:
create thread block level write throttle.  must merge after mv-imm-thread.


chardan:
Okay, typo corrected, Matthew explained why the function's not pure virtual. +1!

PR #34, opened 2012-07-18
Mv badblock recovery

matthewvon:
Rewrite to make BLOCKS.bad file only appear upon first use and only be open when actively saving off a bad block.


jonmeredith:
+1 merge

PR #36, opened 2012-07-30
Mv cache corruption

matthewvon:
This addresses a poor refactoring of Unref() in cache.cc (by me).  Debug / Defense code is left in place in case the bug is just better hidden instead of fixed.


jonmeredith:
Code makes sense and resolves crashers for me.  We'll have to deal with compiler portability issues later.

jonmeredith:
+1 merge.

PR #45, opened 2012-08-16
Mv bloom2

matthewvon:
Bloom filter updated to use prime bit lengths and second hash algorithm.  Includes code to auto select correct bloom filter upon .sst open.


chardan:
+1.

PR #51, opened 2012-09-14
Mv sst stats

matthewvon:
Two items in this check-in:  each .sst file now contains a statistics block and tools/sst_scan.cc added.  "make tools" will compile the sst_scan tool.

### DIFF
--- a/include/leveldb/perf_count.h
+++ b/include/leveldb/perf_count.h
@@ -1,0 +1,256 @@
+// -------------------------------------------------------------------
+//
+// perf_count.h:  performance counters LevelDB (http://code.google.com/p/leveldb/)
+//
+// Copyright (c) 2012 Basho Technologies, Inc. All Rights Reserved.
+//
+// This file is provided to you under the Apache License,
+// Version 2.0 (the "License"); you may not use this file
+// except in compliance with the License.  You may obtain
+// a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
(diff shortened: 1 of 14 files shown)


tools/sst_scan.cc
```diff
@@ -0,0 +1,311 @@
...
+                    total=0;
+                    count=0;
+                    count2=0;
+                    tot_size=0;
+
+                    table = reinterpret_cast<leveldb::TableAndFile*>(table_cache->TEST_GetInternalCache()->Value(handle))->table;
+                    file = reinterpret_cast<leveldb::TableAndFile*>(table_cache->TEST_GetInternalCache()->Value(handle))->file;
+                    it = table->TEST_GetIndexBlock()->NewIterator(options.comparator);
+
+//                if (NULL!=table->FilterObject())
```
chardan: Dead code.

tools/sst_scan.cc
```diff
@@ -0,0 +1,309 @@
...
+    block_info=false;
+    counter_info=false;
+    index_keys=false;
+    csv_header=false;
+    all_keys=false;
+    no_csv=false;
+    counter=0;
+    error_counter=0;
+
+
```
chardan, line 43: Unusual, but confirmed with Matthew that it's NaB. ("./sst_stats -opt1 file -opt2" results in opt2 getting silently ignored. However, the intent is to support "./sst_stats -opt1 file -opt2 file2".)

include/leveldb/perf_count.h
```diff
@@ -0,0 +1,211 @@
...
+extern struct PerformanceCounters * gPerfCounters;
+
+struct PerformanceCounters
+{
+    enum
+    {
+        eVersion=1,            //!< structure versioning
+    };
+
+
```
chardan, line 107: It might be worthwhile "in the distant future" to explore making most of these enumerations that index an array, since most of them share the same type. It depends on how we're using it, but in this code I think there's a lot that would get quite a bit smaller and probably a bit easier to maintain.

chardan:
+1.

PR #52, opened 2012-09-24
Asynchronous write buffers to improve compaction by 14 to to 19%

matthewvon:
This only improves performance in limited cases.  Not merging until it shows benefit in most or all cases.  March 2013 the limit appears to be in number of buffers between read and write side.

December 2013:  This code depends heavily upon passing data between two threads.  Subsequent work demonstrated that thread changes are extremely expensive within the Erlang VM.  Also the mv-hot-threads2 branch is now on master and gives better all round throughput for background operations.  Retiring this pull request.  Leaving the branch since it has a couple of useful memory mapped objects that could be useful for future non-blocking Write designs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,8 @@ TESTS = \
 	write_batch_test
 
 TOOLS = \
-	sst_scan
+	sst_scan \
+	bloom_scan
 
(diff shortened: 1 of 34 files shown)


matthewvon:
Closed without merge/commit.  See comments in PR header.

PR #54, opened 2012-10-18
Fix bug where one bloom filter may not load from .so since not reference...

matthewvon:
...d.  This keeps bloom logic from knowing when/how to decode alternate filters.  Without this fix, alternate bloom filters will be ignored and table files (.sst) containing them will require the full reads to see if keys exist or not, i.e. performance loss but not data loss.


chardan:
+1.

PR #56, opened 2012-10-19
Mv hardware crc

matthewvon:
Dynamically select hardware crc if available ... but only on x86_64 intel processors.


chardan:
+1 (verified: hardware CRC on Intel Core i7; software on AMD Phenom II X4 (no SSE2_4, so this is good) and also VM w/o SSE2_4 (also good))

PR #57, opened 2012-11-05
Mv level dirs

matthewvon:
Code to separate level table files into directories by "level".

### DIFF
--- a/db/repair.cc
+++ b/db/repair.cc
@@ -233,6 +233,7 @@ class Repairer {
     // since ExtractMetaData() will also generate edits.
     FileMetaData meta;
     meta.number = next_file_number_++;
+    meta.level = 0;
     Iterator* iter = mem->NewIterator();
     status = BuildTable(dbname_, env_, options_, table_cache_, iter, &meta, kMaxSequenceNumber);
     delete iter;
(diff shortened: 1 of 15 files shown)


db/repair.cc
```diff
@@ -369,21 +439,33 @@ class Repairer {
...
-  void ArchiveFile(const std::string& fname) {
+  void ArchiveFile(const std::string& fname, bool two_levels=false) {
     // Move into another directory.  E.g., for
     //    dir/foo
     // rename to
     //    dir/lost/foo
-    const char* slash = strrchr(fname.c_str(), '/');
+    std::string::size_type slash, slash2;
+
+    slash=fname.rfind('/');
```
gburd, line 449: rfind always scares me a bit, is there a better way?  What's going on here?

tools/repair.cc
```diff
@@ -0,0 +1,99 @@
...
+
+    return( error_seen && 0!=error_counter ? 1 : 0 );
+
+}   // main
+
+
+void
+command_help()
+{
+    fprintf(stderr, "bloom_scan [option | data_base]*\n");
```
gburd: I the name of this tool "bloom_scan"?

gburd:
Everything but the conversion routine was reviewed successfully.  MvM is going to add some more logic to manage the repair process in cases where the migration process fails mid-way.  Once that's reviewed this is ready to go.

gburd:
+1 to merge

matthewvon:
Thank you.

You in the mood to do one more.  Andy does not seem to be getting around to the mv-perf-counters.  I can do a call on it pretty much anytime.  This is my last contribution to 1.3.

Matthew

PR #60, opened 2012-11-20
hack to counter unexplained condition where threads lack signals

matthewvon:
Previous code would only signal compaction threads if work queues were busy.  The assumption was that the new work would be automatically picked up as threads returned from current work.  Only needed to signal if threads were not working and therefore sitting in a condition wait.

This change has every background work request send a signal, no matter what.  This would have activated the sleeping threads in Ryan's failure case.


chardan:
+1 (no way to directly test the difference, but appears to work ok)

PR #61, opened 2012-11-29
throttle version 3, working against 750 million population test

matthewvon:
This revision to the throttle is in direct response to the new async eleveldb NIF and removal of +S 64:64 from vm.args.  Those two changes to the erlang environment removed other throttling that erlang was implicitly passing to leveldb.  Also, two edge case bugs were fixed in the old throttle logic which remains.


matthewvon:
This check-in / branch has been slowly moved to other branches.  It is no longer needed.

PR #62, opened 2012-12-04
throttle/compaction change that prioritizes the level-0 file compactions...

matthewvon:
... based upon number of files waiting.  This code is a subset of the changes within mv-throttle-3.  Only taking the safest, most common edge case fix.


chardan:
+1

PR #65, opened 2012-12-19
Mv iterator stats:  add performance counters for iterators

matthewvon:
This adds leveldb iterator counters AND support for eleveldb work counters


chardan:
+1

PR #66, opened 2013-01-01
clean up of all warnings but printf lld

matthewvon:


matthewvon:
Merging without code review per mumble discussion.  No code changes.  Just making compiler happy.

PR #67, opened 2013-01-18
Mv throttle 3b

matthewvon:
"final" tuning parameters for 1.3 release are now set.  Review and ship time.


chardan:
+1 to mv_throttle_3b (the only important question is  f4637e1 util/perf_count.cc:L366)

PR #68, opened 2013-01-25
Making counters 32 bits long on 32 bit platforms

engelsanchez:
@matthewvon this is my proposed change to support compilation on 32 bit platforms. As I understand it after talking to Jon, we're not going crazy supporting everything here. This should at least allow leveldb to compile and pass our test suite.  Let me know what you think.


matthewvon:
this was a good solution, but used different approach that kept the counters as 64bits on all platforms.

PR #70, opened 2013-01-25
Mv 32bit counters

matthewvon:
Not as nice as Engel's solution, but keeps counters at 64 bits across all platforms.


engelsanchez:
Well, I'm good with this. I compiled on Ubuntu32 and ran some leveldb tests just in case too. :+1: and I'll leave to you whether you think there's a more portable way to check the wordsize. I think yours will work on anything with glibc anyway.

PR #76, opened 2013-04-05
Mv level work2

matthewvon:
This branch provides further tuning to the mv-level-work1 branch.  The tuning is focused upon reducing when write throttle penalties apply and the performance between levels 2 and 3.  The level 2 to 3 transition is when the code shifts from overlapped .sst files to sorted.  The change slightly increases the write amplification (bad) but reduces latencies, both overall and worst case (good).

### DIFF
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -49,7 +49,7 @@ static struct
     {10485760,  262144000,  576716800,      209715200, 300000000, true},
     {10485760,  262144000,  576716800,      419430400,1500000000, true},
     {10485760,  262144000,  576716800,     4194304000, 314572800, true},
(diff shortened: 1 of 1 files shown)


engelsanchez:
I've reviewed the code enough to understand what is going  on and visually detect any fat fingering and such.

Besides the new tuning magic numbers, we have:
- Penalties not triggering so soon, especially when just reaching # file compaction threshold
- The previous changes to allow levels beyond 0 to have overlapping files was not been completely effective because the code in Compaction:ShouldStopBefore was triggering and smaller compaction files were generated, which ended up hurting performance when compaction through the first non-overlapping level occurred

:+1: :dancer:

PR #77, opened 2013-04-07
Mv cache revert

matthewvon:
This code reverts to Google's original cache implementation.  The code for cache.cc is used twice, once for a file cache and once for a block cache.  I had previously made changes that  yielded nice, incremental performance improvements to the block cache.  No real attempt made to check impact to file cache.  Recent testing shows a file cache miss has 10x or larger impact to performance compared to block cache miss.  The original Google code performs much better for the file cache.  Hence I am reverting to Google's code.

There is one small adjustment to the original LRUCache::Insert routine.  Google's code tests for "space" in the cache upon insertion of new object.  The code will delete the least recently used cache items until enough space is recovered.  This logic fails if the least recently used item is active.  The measure of available space does not change when deleting an active cache item.  The code then loops to delete  the next least recently used item.  The loop can end up deleting a large section of the file cache if the files are active in iterators.  My adjustment only deletes non-active cache items.  This adjustment was previously shipped in my cache replacement code and believed stable / tested.

The branch contains a second adjustment that changes how the file cache size is measured.  The Google code measures files by count.  This makes memory sizing difficult since each file's index block and bloom filter block can vary in size (and I have other changes that vary the size of files by level).  The code now sets the file cache's capacity to 4Mbytes times the max_file_count.  4Mbytes is a median value of samples reviewed from different levels and different datasets.  Each file added to the cache subtracts its actual memory usage.  So the number of files that actually fit in the cache can vary depending upon how file indexes and file bloom filters vary.  But the ability to accurately set the memory model is greatly enhanced / simplified.


matthewvon:
Unfortunately I test things.  Yes, I sometimes even submit pull requests before the multi-hour (multi-day) / multiple configurations stuff finishes if the runtime stuff is meeting expectations.  In this case the Google cache fails after a few hours if the app.config has a small number for max_open_files (50).  The system gets into a thrashing state of opening and closing files.  Yep, I fixed this once before and have put that fix back into the Google cache.cc

The fix is to create a second flavor of cache.  It is initialized via NewLRUCache2().  This second flavor does NOT have an intermediary sharding layer.  The shards are based upon a hash algorithm that does not provide a very even distribution for the file cache.  This leads to lots of cache evictions, thus lots of file opens / closes.  Restoring my previous hack to make the file cache one big happy shard fixes all things.  There might exist a magic capacity value where the Google sharding would perform better for the file cache ... but that takes time and interest.  Passing on both for now.

engelsanchez:
@matthewvon I think the new code in LRUCache::Insert will stop looping as soon as it finds an "old" entry that is still active, and it won't try to find further less "old" entries that could be removed. The one_removed flag is set to false right before checking for active, and used in the while loop to exit. The old code was careful to scan up to the point where there was an inactive entry before running that piece of logic.  Can you check and see if I'm mistaken?

matthewvon:
@engelsanchez Wow.  Good catch.  The intended design was for the while() loop to walk the list of cache items until enough had been deleted to satisfy memory constraints.  Google's original implementation assumed the head of the list was always deleted.  That would cause an infinite loop when I added the if() clause to only delete non-active items.  The one_removed flag stopped the infinite loop.

HOWEVER, your query forced me to really look at Google's intention versus my intention.  My tweak of Google's code was safe, but would never test any object but the head object.  I have submitted new code that will actually walk the list.  Like before, please review for code correctness while I execute the 48 to 72 hours of test to validate.

engelsanchez:
Ok. I've finished my visual inspection of all the code changes. It seems correct as far as I can tell. Let's see what the test harvest brings!

:+1: :dancer:

PR #78, opened 2013-05-09
Repair updated for edge case created with new directory structure.

matthewvon:
See discussion here:  https://github.com/basho/leveldb/wiki/mv-repair-by-level


db/repair.cc
```diff
@@ -115,6 +116,27 @@ class Repairer {
...
+    // perform Riak specific scan for overlapping .sst files
+    //  within a level
+    if (status.ok())
+    {
+        leveldb::DB * db_ptr;
+        Options options;
+
+        db_ptr=NULL;
+        options=org_options_;
+        options.block_cache=options_.block_cache;
```
engelsanchez: Is there any danger in reusing this block cache for the level repair operation here? I haven't fully grasped what the code before this may put in it or where the repair may pull data out of it, but I was wondering if it was safer to just use a fresh one here.

matthewvon: The block cache is global to the database (vnode).  It's pointer has to be passed around different operations simply to give access.  That is what you are seeing here.  Good question though.

engelsanchez:
After talking to @matthewvon about this we have agreed to avoid block cache reuse here to avoid any potential current or future weirdness if semantics change.  In this case, it is only reused within the repair operation. When used from eleveldb, it is not sharing an existing block cache anyway (so not shared with the Riak vnode). So I'm giving this a :+1: :dancer: :boat: conditional on the code changing to create a new block cache (options block_cache set to NULL). It will be taken away if any other naughtiness is introduced, however :)

PR #79, opened 2013-05-17
Mv filecache tuning2

matthewvon:
Tuning to file/table cache to improve random read and iterator performance.  See https://github.com/basho/leveldb/wiki/Mv-filecache-tuning2

About 50% improvement in 2i_slf_stressmix test, with 5x reduction in timeouts from long iterations.


include/leveldb/cache.h
```diff
@@ -90,6 +90,11 @@ class Cache {
   // Allows more accurate tracking of "charge" against each cache item.
   virtual size_t EntryOverheadSize() {return(0);};
 
+  // Riak specific:  Add a reference to cache object to help hold it
+  //  hold it in memory
```
evanmcc: comment nitpick: 'hold it' is repeated.

matthewvon: corrected and checked in.

evanmcc:
Jessie and I have been looking at some breakages in the test suite.  There's a bug in the repair stuff because evict now takes two arguments, but regardless of what you set it to, it seems to introduce a non-deterministic double-release problem.  It's possible to not explode when you do the second release, but the fact that it's started to happen now is worrying.   

Also, master does not pass all of its tests, but haven't looked into that.

Jessie's take:
"Two issues: 1) tool needs patching because of API change; 2) this bug (workaround possible, but concern is that because of the way the assertions are used it might be The Wrong Way To Do It(TM))."

chardan:
+1 (Matthew notes known issues in db_test and corruption_test that he will address later).

PR #80, opened 2013-05-29
I thought I had merged this fix.  Guess I decided not to at the time ......

matthewvon:
... but now it is needed for SmartOS ... which used to not need it.


matthewvon:
Scott reviewed the code and gave +1 via HipChat.

PR #81, opened 2013-06-12
Mv bloom size limit

matthewvon:
See https://github.com/basho/leveldb/wiki/Mv-bloom-size-limit

### DIFF
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -1546,7 +1546,7 @@ bool Compaction::IsBaseLevelForKey(const Slice& user_key) {
     return ret_flag;
 }
 
-bool Compaction::ShouldStopBefore(const Slice& internal_key) {
+bool Compaction::ShouldStopBefore(const Slice& internal_key, size_t key_count) {
 #if 0
     // 04/11/2013 code seems to create way too many small files
     //   in lower levels once highers start to populate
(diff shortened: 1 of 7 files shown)


db/version_set.cc
```diff
@@ -544,19 +551,19 @@ void Version::GetOverlappingInputs(
...
         {
             const std::vector<FileMetaData*>& files = files_[level];
-            int inner, outer;
+            size_t inner, outer;
 
             for (outer=0; outer<files.size()-1 && !overlap_found; ++outer)
             {
                 FileMetaData* outer_meta = files_[level][outer];
-                const Slice outer_start = outer_meta->smallest.user_key();
+                //const Slice outer_start = outer_meta->smallest.user_key();
```
gburd: Any reason not to just delete this code rather than leave it commented out?  If there's a reason to keep it around, maybe the comment should make note of that.

db/version_set.cc
```diff
@@ -544,19 +551,19 @@ void Version::GetOverlappingInputs(
...
                 const Slice outer_limit = outer_meta->largest.user_key();
 
                 for (inner=outer+1; inner<files.size() && !overlap_found; ++inner)
                 {
                     FileMetaData* inner_meta = files_[level][inner];
                     const Slice inner_start = inner_meta->smallest.user_key();
-                    const Slice inner_limit = inner_meta->largest.user_key();
+                    //const Slice inner_limit = inner_meta->largest.user_key();
```
gburd: Same comment here as for line 552, why keep this around?

db/version_set.cc
matthewvon: Laziness.  Will remove before push to master.

db/version_set.cc
matthewvon: Laziness.  Will remove before push to master.

gburd:
After reviewing and testing the code in the small (on my laptop) and then discussing with MvM these new heuristics for compaction and which levels may contain overlapping sst files I feel that these changes will improve things as anticipated and I don't see any obvious drawbacks.  +1 to merge

PR #83, opened 2013-06-19
create new option specifically for compaction CRC test control

matthewvon:
compaction CRC test was previously one of multiple tests controlled by the paranoid_checks option variable.  We do not want all paranoid_checks, only compaction CRC.  This new variable gives us that control.

### DIFF
--- a/db/c.cc
+++ b/db/c.cc
@@ -416,6 +416,11 @@ void leveldb_options_set_paranoid_checks(
   opt->rep.paranoid_checks = v;
 }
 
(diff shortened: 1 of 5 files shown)


engelsanchez:
The code looks good to me.  I couldn't find any other place that needed change, it passes the internal tests.
:+1:

engelsanchez:
Just one small thing: The C api would need a new function to set the option (file c.cc) . But we don't use that, so...

PR #86, opened 2013-07-10
Mv riak 1 4 cleanup

matthewvon:
Move Basho's throttle code to independent throttle.cc and throttle.h.  Was previously hacked into the env_posix.cc source file.  The throttle code is now ready for its own home.


bookshelfdave:
looks fine, compiles
+1

PR #89, opened 2013-08-02
Mv level work4

matthewvon:
4 small changes:  details here https://github.com/basho/leveldb/wiki/Mv-level-work4

### DIFF
--- a/db/version_set.h
+++ b/db/version_set.h
@@ -107,7 +107,7 @@ class Version {
 
   const std::vector<FileMetaData*> & GetFileList(int level) const {return files_[level];};
 
-  int WritePenalty() const {return write_penalty_; }
+  volatile int WritePenalty() const {return write_penalty_; }
 
   // Riak specific repair routine
   bool VerifyLevels(int & level, InternalKey & begin, InternalKey & end);
@@ -139,7 +139,7 @@ class Version {
   // are initialized by Finalize().
   double compaction_score_;
   int compaction_level_;
-  int write_penalty_;
+  volatile int write_penalty_;
 
   explicit Version(VersionSet* vset)
       : vset_(vset), next_(this), prev_(this), refs_(0),
--- a/util/env_posix.cc
+++ b/util/env_posix.cc
@@ -119,7 +119,7 @@ class PosixRandomAccessFile: public RandomAccessFile {
     // Currently hurts performance instead of helps.  Likely
     //  requires better interaction with tables already in cache
     //  that start compaction.  See comment in table_cache.cc.
-    //    posix_fadvise(fd_, 0, file_size_, POSIX_FADV_RANDOM);
+    posix_fadvise(fd_, 0, file_size_, POSIX_FADV_RANDOM);
 #endif
   }
   virtual ~PosixRandomAccessFile()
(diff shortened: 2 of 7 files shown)


db/version_set.h
```diff
@@ -139,7 +139,7 @@ class Version {
   // are initialized by Finalize().
   double compaction_score_;
   int compaction_level_;
-  int write_penalty_;
+  volatile int write_penalty_;
```
gburd, line 142: I'm not sure what behavior you're trying to get with the addition of "volatile" as it doesn't prevent data race conditions or multi-instruction loads so frequently it doesn't accomplish what you'd think in multi-core/CPU-threaded code.  So, what's your intention here? (references: http://software.intel.com/en-us/blogs/2007/11/30/volatile-almost-useless-for-multi-threaded-programming http://stackoverflow.com/questions/8819095/concurrency-atomic-and-volatile-in-c11-memory-model/8833218#8833218 http://bartoszmilewski.com/2008/12/01/c-atomics-and-memory-ordering/)

util/env_posix.cc
```diff
@@ -929,9 +930,15 @@ class PosixEnv : public Env {
     else
     {
         // low priority compaction (not imm, not level0)
+#if 1
```
gburd: Do you need to keep this if/else/endif around for some reason?  Seems to me to be just a reminder which could be a comment or simply left to the file's history in git.

util/env_posix.cc
```diff
@@ -944,6 +951,30 @@ class PosixEnv : public Env {
  * support direct access to underlying type ... which is needed.
  */
 void
+PosixEnv::InsertQueue0(
+    PosixEnv::BGItem & item)
+{
+    BGQueue::iterator it;
+    bool looking;
+
+    // this is slow and painful with deque as underlying container
```
gburd, line 954: Indeed that does look slow/painful.  Looks like a priority queue insert into a dequeue data structure.  Does this happen on a path that's used frequently enough to bother looking for a better implementation?

gburd:
With the exception of the confusing use of "volatile" I'm +1 for merge on this set of changes.  Even that seems minor given where it's used, but I'd like to clarify that first.

db/version_set.h
matthewvon, line 142: (software.intel.com link worked yesterday, and failed today, so was not reread before writing this comment.  I think it has comments about what I am saying here.  The other two are not as relevant.)  Unless I made a mistake, no critical changes are ever made to a variable with volatile access without proper synchronization methods.  My intended usage of volatile is two fold.  First I like to use it to highlight variables to the next programmer that are referenced between threads independent of the the entire class / struct.  I may or may not have proper mutex or atomic synchronization on the variable depending upon its criticality.  Second I use volatile in hopes of limiting unexpected compiler optimizations that will change the code logic.  I am hinting to the compiler that the value should be reread.  The later is important if my intention is to reference the value of a variable that may be changing in another thread ... and make the reference without going through the expense of atomics / mutexes.

util/env_posix.cc
matthewvon, line 954: The intent is to remove both InsertQueue0 and InsertQueue2.  The replacement will be a Set or Map which will keep contents in sorted order.  This kludge is to allow me to keep original Google container without a big rewrite.  The big rewrite is now overdue.  I have high hopes that is going to happen for Riak 2.0.

util/env_posix.cc
matthewvon: Oops ... might be gone by the time you check the code again.

db/version_set.h
gburd, line 142: In my experience and from what I've read the volatile keyword does more harm than good in most cases.  A comment could be used to indicate the intended use across threads.  But, I think this falls into the category of YMMV and isn't a critical concern.

gburd:
+1 to merge, nice work

PR #90, opened 2013-08-09
leveldb's mv-clean-overlaps

matthewvon:
overlap files do NOT properly clean up memory on database close.

Details here:  https://github.com/basho/leveldb/wiki/Mv-clean-overlaps


jaredmorrow:
+1, looks right and tests pass.

PR #92, opened 2013-08-22
Mv spin locks

matthewvon:
Details of the changes are here:  https://github.com/basho/leveldb/wiki/Mv-spin-locks

### DIFF
--- a/db/db_impl.cc
+++ b/db/db_impl.cc
@@ -1564,7 +1564,7 @@ Status DBImpl::Write(const WriteOptions& options, WriteBatch* my_batch) {
       // throttle is per key write, how many in batch?
       count=(NULL!=my_batch ? WriteBatchInternal::Count(my_batch) : 1);
       env_->SleepForMicroseconds(throttle * count);
-      gPerfCounters->Add(ePerfDebug0, throttle);
+      gPerfCounters->Add(ePerfDebug0, throttle * count);
   }   // if
 
   return status;
--- a/port/port_posix.h
+++ b/port/port_posix.h
@@ -103,7 +103,7 @@ class Mutex {
 };
 
 
-#if defined(_POSIX_SPIN_LOCKS)
+#if defined(_POSIX_SPIN_LOCKS) && 0<_POSIX_SPIN_LOCKS
 class Spin {
  public:
   Spin();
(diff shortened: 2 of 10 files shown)


bookshelfdave:
@matthewvon looks like this needs a rebase or some cleanup (from "We can’t automatically merge this pull request."). I can do some review tomorrow.

matthewvon:
The issue is that I manually added .travis.yml to both this branch and another branch already on master.  The two files differed by a line feed at the end ... git is so wise.

db/db_impl.cc
```diff
@@ -1564,7 +1564,7 @@ Status DBImpl::Write(const WriteOptions& options, WriteBatch* my_batch) {
       // throttle is per key write, how many in batch?
       count=(NULL!=my_batch ? WriteBatchInternal::Count(my_batch) : 1);
       env_->SleepForMicroseconds(throttle * count);
-      gPerfCounters->Add(ePerfDebug0, throttle);
+      gPerfCounters->Add(ePerfDebug0, throttle * count);
```
bookshelfdave, line 1567: is this related to spin locks, or just a perf counter bug fix?

port/port_posix.h
```diff
@@ -102,6 +102,30 @@ class Mutex {
...
+
+#if defined(_POSIX_SPIN_LOCKS) && 0<_POSIX_SPIN_LOCKS
+class Spin {
+ public:
+  Spin();
+  ~Spin();
+
+  void Lock();
+  void Unlock();
+  void AssertHeld() { }
```
bookshelfdave, line 114: I see AssertHeld() { } in other classes, what's the point?

bookshelfdave:
+1

all tests pass with `make check`

db/db_impl.cc
matthewvon, line 1567: unrelated bug fix ... saw it was wrong, fixed it.

port/port_posix.h
matthewvon, line 114: I copied the Mutex class, then changed the names to Spin.  I need spin to be 100% compatible with Mutex ... since 1) I am replacing Mutex with Spin in several places and 2) Spin becomes Mutex on platforms that do not have pthread_spinlock_t.

PR #94, opened 2013-08-31
branch mv-valgrind-cleanup

matthewvon:
Branch discussion is here:  https://github.com/basho/leveldb/wiki/mv-valgrind-cleanup

### DIFF
--- a/db/repair.cc
+++ b/db/repair.cc
@@ -61,13 +61,24 @@ class Repairer {
   }
 
   ~Repairer() {
-    delete table_cache_;
     if (owns_info_log_) {
       delete options_.info_log;
     }
     if (owns_cache_) {
       delete options_.block_cache;
     }
+
+    // must remove second ref counter that keeps overlapped files locked
+    //  table cache
+    for (int level = 0; level < config::kNumLevels; level++) {
+        {
+            for (size_t i = 0; i < table_numbers_[level].size(); i++) {
+                table_cache_->Evict(table_numbers_[level][i], (level<leveldb::config::kNumOverlapLevels));
+            }   // for
+        }   // if
+    } // for
+
+    delete table_cache_;
   }
 
   Status Run() {
--- a/db/repair.cc
+++ b/db/repair.cc
@@ -70,10 +70,12 @@ class Repairer {
 
     // must remove second ref counter that keeps overlapped files locked
     //  table cache
+    bool is_overlap;
     for (int level = 0; level < config::kNumLevels; level++) {
         {
+            is_overlap=(level < leveldb::config::kNumOverlapLevels);
             for (size_t i = 0; i < table_numbers_[level].size(); i++) {
-                table_cache_->Evict(table_numbers_[level][i], (level<leveldb::config::kNumOverlapLevels));
+                table_cache_->Evict(table_numbers_[level][i], is_overlap);
             }   // for
         }   // if
     } // for
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -821,17 +821,6 @@ VersionSet::~VersionSet() {
   // must remove second ref counter that keeps overlapped files locked
   //  table cache
 
-#if 0
-  for (int level = 0; level < config::kNumLevels; level++) {
-    if (gLevelTraits[level].m_OverlappedFiles)
-    {
-        for (size_t i = 0; i < current_->NumFiles(level); i++) {
-            FileMetaData* f = current_->files_[level][i];
-            table_cache_->Evict(f->number, true);
-        }   // for
-    }   // if
-  } // for
-#endif
   current_->Unref();
   assert(dummy_versions_.next_ == &dummy_versions_);  // List must be empty
   delete descriptor_log_;
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -100,11 +100,11 @@ Version::~Version() {
       assert(f->refs > 0);
       f->refs--;
 
-      // clear Riak's double reference of overlapped files
-      if (vset_->IsLevelOverlapped(level))
+      if (f->refs <= 0) {
+        // clear Riak's double reference of overlapped files
+        if (vset_->IsLevelOverlapped(level))
           vset_->GetTableCache()->Evict(f->number, true);
 
-      if (f->refs <= 0) {
         delete f;
       }
     }
@@ -821,6 +821,7 @@ VersionSet::~VersionSet() {
   // must remove second ref counter that keeps overlapped files locked
   //  table cache
 
+#if 0
   for (int level = 0; level < config::kNumLevels; level++) {
     if (gLevelTraits[level].m_OverlappedFiles)
     {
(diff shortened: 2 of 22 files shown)


db/repair.cc
```diff
@@ -61,13 +61,24 @@ class Repairer {
...
     if (owns_cache_) {
       delete options_.block_cache;
     }
+
+    // must remove second ref counter that keeps overlapped files locked
+    //  table cache
+    for (int level = 0; level < config::kNumLevels; level++) {
+        {
+            for (size_t i = 0; i < table_numbers_[level].size(); i++) {
+                table_cache_->Evict(table_numbers_[level][i], (level<leveldb::config::kNumOverlapLevels));
```
bookshelfdave: I'm not really a fan of the (level<leveldb::config::kNumOverlapLevels) parameter. Can you split it out for clarity?

db/version_set.cc
```diff
@@ -813,6 +818,9 @@ class VersionSet::Builder {
 }
 
 VersionSet::~VersionSet() {
+  // must remove second ref counter that keeps overlapped files locked
```
bookshelfdave, line 821: is this a TODO, or a description for current_->Unref()?

matthewvon, line 821: description

include/leveldb/db.h
```diff
@@ -14,7 +14,7 @@
 
 // Update Makefile if you change these
 static const int kMajorVersion = 1;
-static const int kMinorVersion = 5;
+static const int kMinorVersion = 9;
```
bookshelfdave, line 17: Makefile is in sync. Just checkin'.

db/repair.cc
matthewvon: changed

include/leveldb/db.h
matthewvon, line 17: they were out of sync.  fixed.  not that anything really cares.

util/cache.cc
```diff
@@ -211,9 +211,10 @@ class LRUCache : public Cache {
   for (LRUHandle* e = lru_.next; e != &lru_; ) {
     LRUHandle* next = e->next;
 
+    assert(e->refs == 1);  // Error if caller has an unreleased handle
```
bookshelfdave, line 214: Can you update the comment to explain why you don't need `|| e->refs == 2`?

util/env_posix.cc
```diff
@@ -783,6 +782,7 @@ class PosixEnv : public Env {
 PosixEnv::PosixEnv() : page_size_(getpagesize()),
                        started_bgthread_(false),
                        bg_backlog_(0), bg_active_(0),
+                       bgthread_running_(true), bgthread_count_(0),
```
bookshelfdave, line 785: Shouldn't these be specified in declaration order? (per Effective C++)

util/env_posix.cc
```diff
@@ -798,6 +798,34 @@ class PosixEnv : public Env {
...
+    // be sure thread actually started before attempting
+    //  shutdown and join.
+    if (0!=bgthread_count_)
+    {
+        while(0!=bgthread_count_)
+        {
+            PthreadCall("lock", pthread_mutex_lock(&mu_));
+            PthreadCall("broadcast", pthread_cond_broadcast(&bgsignal_));
+            PthreadCall("unlock", pthread_mutex_unlock(&mu_));
+            SleepForMicroseconds(10);
```
bookshelfdave, line 816: why sleep?

util/env_posix.cc
```diff
@@ -798,6 +798,34 @@ class PosixEnv : public Env {
...
 
+
+PosixEnv::~PosixEnv()
+{
+    // kill off background threads
+    bgthread_running_=false;
+
+    // be sure thread actually started before attempting
+    //  shutdown and join.
+    if (0!=bgthread_count_)
```
bookshelfdave, line 809: could there be a race here?

util/cache.cc
matthewvon, line 214: e->refs == 1 is the original Google code.  I made a mistake in believing cache entries were still cleanable with e->refs ==2 (due to my intentional extra ref on overlapped files), and was then forcing the cleanup.  Bad programming.  Getting the unit tests running again showed me the error of my ways.

util/env_posix.cc
```diff
@@ -1026,29 +1062,38 @@ void PosixEnv::BGThread()
...
-            gPerfCounters->Inc(ePerfBGCompactImm);
-        else if (bgthread3_==pthread_self())
-            gPerfCounters->Inc(ePerfBGCloseUnmap);
-        else if (bgthread2_==pthread_self())
-            gPerfCounters->Inc(ePerfBGCompactLevel0);
+            if (bgthread4_==pthread_self())
+                gPerfCounters->Inc(ePerfBGCompactImm);
+            else if (bgthread3_==pthread_self())
+                gPerfCounters->Inc(ePerfBGCloseUnmap);
+            else if (bgthread2_==pthread_self())
```
bookshelfdave, line 1084: Why not just call pthread_self() once before the if/else block instead of in each clause?

util/env_posix.cc
```diff
@@ -1315,6 +1363,27 @@ Env* Env::Default() {
...
+        for (loop=0; loop<THREAD_BLOCKS; ++loop)
+        {
+            delete default_env[loop];
+            default_env[loop]=NULL;
+        }   // for
+
+        ThrottleShutdown();
+    }   // if
+
+    ComparatorShutdown();
```
bookshelfdave, line 1382: just checking that this doesn't need to be part of

```
if (started)
    { ... }
```

util/env_posix.cc
matthewvon, line 785: uh ... not seeing this one ... which do you see as out of order?  but I do see that leveldb's compiler defaults do not warn if the order is wrong ... hmm

util/env_posix.cc
matthewvon, line 816: yield() might work ... but compactions could be active so I do not want to poll too often.  we could be in this loop for a while awating the compaction completion.

util/env_posix.cc
bookshelfdave, line 785: Sorry, I should have commented on a lower line.
clock_res_(1) appears out of order.

util/env_posix.cc
matthewvon, line 1084: This routine was modified, modified, and remodified in a series of hacks.  It is a kludge waiting to fall apart.  The goal is to rewrite the entire threading logic in Riak 2.0 based upon the now known characteristics of compaction work loads.  New work will use "hot threads" thread pool of eleveldb to create 4 pools (with far better management code than what you see).  Pools will be universal, not fixed count per env object.

util/env_posix.cc
matthewvon, line 1382: huh?  You asking about ComparatorShutdown?  If so, no.  The resource managed there now has a NULL default so the Shutdown in that routine has its own control mechanism.

bookshelfdave:
+1
test results: https://gist.github.com/metadave/6438077

util/env_posix.cc
matthewvon, line 809: turns out there was a race in BGThread related to these two variables.  Good question.

matthewvon:
manual merge completed.

PR #95, opened 2013-09-24
Mv flexcache

matthewvon:
see details here https://github.com/basho/leveldb/wiki/mv-flexcache

When used with Riak, needs mv-flexcache branch of basho/eleveldb also.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,7 @@ TESTS = \
 	env_test \
 	filename_test \
 	filter_block_test \
+	flexcache_test \
 	log_test \
 	memenv_test \
 	skiplist_test \
@@ -152,6 +153,9 @@ filename_test: db/filename_test.o $(LIBOBJECTS) $(TESTHARNESS)
 
 filter_block_test: table/filter_block_test.o $(LIBOBJECTS) $(TESTHARNESS)
 	$(CXX) table/filter_block_test.o $(LIBOBJECTS) $(TESTHARNESS) -o $@ $(LDFLAGS)
+
+flexcache_test: util/flexcache_test.o $(LIBOBJECTS) $(TESTHARNESS)
+	$(CXX) util/flexcache_test.o $(LIBOBJECTS) $(TESTHARNESS) -o $@ $(LDFLAGS)
 
 log_test: db/log_test.o $(LIBOBJECTS) $(TESTHARNESS)
 	$(CXX) db/log_test.o $(LIBOBJECTS) $(TESTHARNESS) -o $@ $(LDFLAGS)
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ TESTS = \
 	bloom_test \
 	c_test \
 	cache_test \
+	cache2_test \
 	coding_test \
 	corruption_test \
 	crc32c_test \
@@ -129,6 +130,9 @@ c_test: db/c_test.o $(LIBOBJECTS) $(TESTHARNESS)
 
 cache_test: util/cache_test.o $(LIBOBJECTS) $(TESTHARNESS)
 	$(CXX) util/cache_test.o $(LIBOBJECTS) $(TESTHARNESS) -o $@ $(LDFLAGS)
+
+cache2_test: util/cache2_test.o $(LIBOBJECTS) $(TESTHARNESS)
+	$(CXX) util/cache2_test.o $(LIBOBJECTS) $(TESTHARNESS) -o $@ $(LDFLAGS)
 
 coding_test: util/coding_test.o $(LIBOBJECTS) $(TESTHARNESS)
 	$(CXX) util/coding_test.o $(LIBOBJECTS) $(TESTHARNESS) -o $@ $(LDFLAGS)
--- a/db/db_impl.cc
+++ b/db/db_impl.cc
@@ -117,12 +117,13 @@ static void ClipToRange(T* ptr, V minvalue, V maxvalue) {
 Options SanitizeOptions(const std::string& dbname,
                         const InternalKeyComparator* icmp,
                         const InternalFilterPolicy* ipolicy,
-                        const Options& src) {
+                        const Options& src,
+                        Cache * block_cache) {
   Options result = src;
   result.comparator = icmp;
   result.filter_policy = (src.filter_policy != NULL) ? ipolicy : NULL;
 //  ClipToRange(&result.max_open_files,            20,     50000);
-//  ClipToRange(&result.write_buffer_size,         64<<10, 1<<30);
+  ClipToRange(&result.write_buffer_size,         64<<10, 1<<30);
   ClipToRange(&result.block_size,                1<<10,  4<<20);
   if (result.info_log == NULL) {
     // Open a log file in the same directory as the db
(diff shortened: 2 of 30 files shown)


Makefile
```diff
@@ -129,6 +131,9 @@ c_test: db/c_test.o $(LIBOBJECTS) $(TESTHARNESS)
 cache_test: util/cache_test.o $(LIBOBJECTS) $(TESTHARNESS)
 	$(CXX) util/cache_test.o $(LIBOBJECTS) $(TESTHARNESS) -o $@ $(LDFLAGS)
 
+cache2_test: util/cache2_test.o $(LIBOBJECTS) $(TESTHARNESS)
+	$(CXX) util/cache2_test.o $(LIBOBJECTS) $(TESTHARNESS) -o $@ $(LDFLAGS)
```
seancribbs, line 135: Running cache2_test, I get this failure:

```
$ ./cache2_test 
==== Test CacheTest.HitAndMiss
==== Test CacheTest.Erase
==== Test CacheTest.EntriesArePinned
==== Test CacheTest.EvictionPolicy
==== Test CacheTest.HeavyEntries
==== Test CacheTest.FlushedEntries
==== Test CacheTest.FileCacheExpire
util/cache2_test.cc:246: failed: 94371840 == 188743680
```

Running OS/X 10.8.5 with the latest XCode (5.0)

util/db_list.h
```diff
@@ -0,0 +1,73 @@
...
+ * DBList:  class to provide management access to all 
+ *  open databases (Riak vnodes)
+ */
+class DBListImpl
+{
+protected:
+   typedef std::set<DBImpl *> db_set_t;
+
+
+public:
```
seancribbs: You have several empty sections here you might consider consolidating.

util/flexcache.cc
```diff
@@ -0,0 +1,130 @@
...
+ */
+uint64_t
+FlexCache::GetDBCacheCapacity(
+    bool IsInternal)   //!< value describing cache attributes of caller
+{
+    uint64_t ret_val, shared_total;
+    size_t count, internal_count;
+
+    // get count of database by type
+    count=DBList()->GetDBCount(IsInternal);
```
seancribbs, line 83: Why doesn't this method just return a pair of normal/internal counts?

Makefile
seancribbs, line 135: This was resolved in c03ce62

db/db_impl.cc
```diff
@@ -194,9 +212,9 @@ Options SanitizeOptions(const std::string& dbname,
   if (owns_info_log_) {
     delete options_.info_log;
   }
-  if (owns_cache_) {
-    delete options_.block_cache;
-  }
+//  if (owns_cache_) {
```
seancribbs: Why is this left commented?

db/db_test.cc
```diff
@@ -1551,7 +1551,7 @@ static bool Between(uint64_t val, uint64_t low, uint64_t high) {
   Options options = CurrentOptions();
   options.env = env_;
   options.block_cache = NewLRUCache(0);  // Prevent cache hits
-  options.filter_policy = NewBloomFilterPolicy(10);
+  options.filter_policy = NewBloomFilterPolicy2(16);
```
seancribbs, line 1554: What does this bloom filter change have to do with the double cache?

db/log_writer.cc
```diff
@@ -74,7 +74,7 @@ Status Writer::AddRecord(const Slice& slice) {
 
 Status Writer::EmitPhysicalRecord(RecordType t, const char* ptr, size_t n) {
   assert(n <= 0xffff);  // Must fit in two bytes
-  assert(block_offset_ + kHeaderSize + n <= kBlockSize);
+  assert(block_offset_ + kHeaderSize + (int)n <= kBlockSize);
```
seancribbs, line 77: I assume this is just a safety change?

include/leveldb/options.h
```diff
@@ -98,7 +99,7 @@ struct Options {
   // Number of open files that can be used by the DB.  You may need to
   // increase this if your database has a large working set (budget
   // one open file per 2MB of working set).
-  //
+  // RIAK: NO LONGER USED
```
seancribbs, line 102: What is the danger of removing this option from the struct?

db/db_impl.cc
```diff
@@ -183,6 +197,10 @@ Options SanitizeOptions(const std::string& dbname,
   }
   mutex_.Unlock();
 
+  // make sure flex cache knows this db is gone
+  //   but we do not know the current size ... give 0
+  gFlexCache.SetTotalMemory(0);
```
seancribbs: Like we discussed in the mumble, setting the total memory size and freeing a database's cache resources seem like separate things, even though they both cause the reclamation of cache space and recomputation of how much each DB is given. Also, `0` as a magic value there seems like it's getting around designing an API that is more transparent. What if instead it looked like this:

``` cpp
gFlexCache.DBOpened();
gFlexCache.DBClosed();
// internally, these both could call ResizeCaches() or whatever
```

Again, those are just suggestions.

gburd:
Matthew, I've read and re-read the code and tried it out locally.  I think this is good to go.  I'd like to run more stress tests, but I think merging early for this kind of change will help shake out integration issues early.  On that note, I'm +1 to merge as soon as you've addressed Sean's comments.

db/db_impl.cc
matthewvon: removed

db/log_writer.cc
matthewvon, line 77: eliminates one compiler warning on a debug build

db/db_test.cc
matthewvon, line 1554: nothing.  but since I changed bloom filters 10 months ago ... I thought I best fix this while I was thinking about it.

seancribbs, line 1554: Seems reasonable :smile:

include/leveldb/options.h
matthewvon, line 102: I am being indecisive about this param.  Just verified that it is no longer in active use.  So only impact would be to community users that have leveldb, but not eleveldb / riak.  Would cause their compile to break on update ... which might be a good thing since it screams "something has changed".  I intentionally did not remove "block_cache" because Google has some special configurations in the unit tests that set the value directly (i.e. not via double_cache).  Open to opinions / direction on max_open_files.

seancribbs, line 102: ^^ @jonmeredith @gburd Thoughts?

db/db_impl.cc
matthewvon: added FlexCache::RecalculateAllocations()

util/db_list.h
matthewvon: done

util/flexcache.cc
matthewvon, line 83: Why?  std::pair is not my typical programming style.  I have therefore run 3 different versions of the function 1 billion times.  Results are rounded to seconds.

```
uint64_t GetDBCacheCapacity(bool IsInternalDB)                        202 seconds
std::pair<uint64_t, uint64_t> GetDBCacheCapacityPair()              268 seconds
void GetDBCacheCapacity(uint64_t & internal, uint64_t & user);   270 seconds
```

(Ran the original function both before the other two and after to verify there was no startup impact.)

I think the key point here is that the caller of this routine only wants one value.  It is passing to the function a flag that was passed to it upon creation.  Returning both values will require more code at the calling points (in cache2.cc) to pick the proper value will further complicate the code.  I recommend staying with the current implementation since it is both faster and fits current usage best.

Issue #87, opened 2013-07-14
db/version_set.cc:60: error: integer constant is too large for ‘long’ type


BillBarnhill:
#  Environment
- Mac OS X 10.7.5
- Using Riak source from http://s3.amazonaws.com/downloads.basho.com/riak/1.4/1.4.0/riak-1.4.0.tar.gz - fetched on 14 JUL 2013
- Erlang R15B01 (erts-5.9.1) [source] [smp:4:4] [async-threads:0] [hipe] [kernel-poll:false]
- $ cc -v
  - Using built-in specs.
  - Target: i686-apple-darwin11
  - Configured with: 
    - /private/var/tmp/llvmgcc42/llvmgcc42-2336.1~22/src/configure - 
    - -disable-checking 
    - --enable-werror 
    - --prefix=/Developer/usr/llvm-gcc-4.2 
    - --mandir=/share/man 
    - --enable-languages=c,objc,c++,obj-c++ 
    - --program-prefix=llvm- 
    - --program-transform-name=/^[cg][^.-]*$/s/$/-4.2/ 
    - --with-slibdir=/usr/lib 
    - --build=i686-apple-darwin11 
    - --enable-llvm=/private/var/tmp/llvmgcc42/llvmgcc42-2336.1~22/dst-llvmCore/Developer/usr/local 
    - --program-prefix=i686-apple-darwin11- 
    - --host=x86_64-apple-darwin11 
    - --target=i686-apple-darwin11 
    - --with-gxx-include-dir=/usr/include/c++/4.2.1
  - Thread model: posix
  - gcc version 4.2.1 (Based on Apple Inc. build 5658) (LLVM build 2336.1.00)
# Expected behavior

All deps build clean.
# Observed behavior

The following error emitted when building eleveldb/leveldb dep:

```
db/version_set.cc:59: warning: this decimal constant is unsigned only in ISO C90
db/version_set.cc:59: warning: this decimal constant is unsigned only in ISO C90
db/version_set.cc:60: error: integer constant is too large for ‘long’ type
db/version_set.cc:60: error: integer constant is too large for ‘long’ type
db/version_set.cc:61: error: integer constant is too large for ‘long’ type
db/version_set.cc:61: error: integer constant is too large for ‘long’ type
db/version_set.cc:62: error: integer constant is too large for ‘long’ type
db/version_set.cc:62: error: integer constant is too large for ‘long’ type
[...]

PR #96, opened 2013-09-24
Fix version_set compilation bug

fgallaire:


matthewvon:
Yes, pull request accepted.  Will merge soon.

What platform / compiler gave warning / error?  I would like to reproduce.

fgallaire:
https://github.com/basho/leveldb/issues/87

matthewvon:
In a perfect world I would be able to find OSX 10.7.x machine with related XCode to verify before / after effects of the change.  In my world that did not happen.  I have verified that the change does NOT break my 10.8 OSX XCode and Linux builds.  Also found C or C++ spec validating the usage ... merging.

matthewvon:
p.s.  I manually made the same change in branch mv-hot-threads1 ... expect a merge conflict there until that branch goes to master also.

PR #97, opened 2013-10-10
1.4.2 turner

matthewvon:
1.  fix race condition in close of async recovery log
2.  address performance problems (write speed and level-0 compaction) with big objects

Note:  spin locks taken from earlier PR already approved for Riak 2.0, and hot_threads taken eleveldb async threads of Riak 1.3.

### DIFF
--- a/db/db_impl.cc
+++ b/db/db_impl.cc
@@ -33,8 +33,10 @@
 #include "table/merger.h"
 #include "table/two_level_iterator.h"
 #include "util/coding.h"
+#include "util/hot_threads.h"
 #include "util/logging.h"
 #include "util/mutexlock.h"
+#include "util/thread_tasks.h"
 #include "util/throttle.h"
 #include "leveldb/perf_count.h"
 
@@ -177,7 +179,7 @@ DBImpl::~DBImpl() {
   // Wait for background work to finish
   mutex_.Lock();
   shutting_down_.Release_Store(this);  // Any non-NULL value is ok
-  while (bg_compaction_scheduled_) {
+  while (IsCompactionScheduled()) {
     bg_cv_.Wait();
   }
   mutex_.Unlock();
@@ -480,7 +482,7 @@ void DBImpl::CheckCompactionState()
         int level;
 
         // wait out executing compaction (Wait gives mutex to compactions)
-        if (bg_compaction_scheduled_)
+        if (IsCompactionScheduled())
             bg_cv_.Wait();
 
         for (level=0, need_compaction=false;
@@ -500,7 +502,7 @@ void DBImpl::CheckCompactionState()
             }   //if
         }   // for
 
-    } while(bg_compaction_scheduled_ && need_compaction);
+    } while(IsCompactionScheduled() && need_compaction);
 
     if (log_flag)
         Log(options_.info_log, "Cleanup compactions completed ... DB::Open continuing");
@@ -610,7 +612,7 @@ Status DBImpl::RecoverLogFile(uint64_t log_number,
   return status;
 }
 
-Status DBImpl::WriteLevel0Table(MemTable* mem, VersionEdit* edit,
+Status DBImpl::WriteLevel0Table(volatile MemTable* mem, VersionEdit* edit,
                                 Version* base) {
   mutex_.AssertHeld();
   const uint64_t start_micros = env_->NowMicros();
@@ -618,7 +620,7 @@ Status DBImpl::WriteLevel0Table(MemTable* mem, VersionEdit* edit,
   meta.number = versions_->NewFileNumber();
   meta.level = 0;
   pending_outputs_.insert(meta.number);
-  Iterator* iter = mem->NewIterator();
+  Iterator* iter = ((MemTable *)mem)->NewIterator();
   SequenceNumber smallest_snapshot;
   Log(options_.info_log, "Level-0 table #%llu: started",
       (unsigned long long) meta.number);
@@ -769,7 +771,7 @@ void DBImpl::TEST_CompactRange(int level, const Slice* begin,const Slice* end) {
 
   MutexLock l(&mutex_);
   while (!manual.done) {
-    while (manual_compaction_ != NULL || bg_compaction_scheduled_) {
+    while (manual_compaction_ != NULL || IsCompactionScheduled()) {
       bg_cv_.Wait();
     }
     manual_compaction_ = &manual;
@@ -836,18 +838,18 @@ void DBImpl::MaybeScheduleCompaction() {
   if (push)
   {
       // is something else already on background work queue
-      if (bg_compaction_scheduled_)
+      if (IsCompactionScheduled())
           state=2;
       else
           state=1;
   }   // if
 
 
-  if (bg_compaction_scheduled_ && !push) {
+  if (IsCompactionScheduled() && !push) {
     // Already scheduled
   } else if (shutting_down_.Acquire_Load()) {
     // DB is being deleted; no more background compactions
-  } else if (imm_ == NULL &&
+  } else if (//imm_ == NULL &&
              manual_compaction_ == NULL &&
              !versions_->NeedsCompaction()) {
     // No work to be done
@@ -863,7 +865,7 @@ void DBImpl::BGWork(void* db) {
 
 void DBImpl::BackgroundCall() {
   MutexLock l(&mutex_);
-  assert(bg_compaction_scheduled_);
+  assert(IsCompactionScheduled());
   if (!shutting_down_.Acquire_Load()) {
     Status s = BackgroundCompaction();
     if (!s.ok()) {
@@ -892,7 +894,7 @@ void DBImpl::BackgroundCall() {
 void 
 DBImpl::BackgroundImmCompactCall() {
   MutexLock l(&mutex_);
-//  assert(bg_compaction_scheduled_);
+//  assert(IsCompactionScheduled());
   if (!shutting_down_.Acquire_Load()) {
     Status s = CompactMemTable();
     if (!s.ok()) {
@@ -908,13 +910,13 @@ DBImpl::BackgroundImmCompactCall() {
       mutex_.Lock();
     }
   }
-//  bg_compaction_scheduled_ = false;
+//  IsCompactionScheduled() = false;
 
   // Previous compaction may have produced too many files in a level,
   // so reschedule another compaction if needed.
   if (!options_.is_repair)
       MaybeScheduleCompaction();
-//  bg_cv_.SignalAll();
+  bg_cv_.SignalAll();
 }
 
 
@@ -924,7 +926,7 @@ Status DBImpl::BackgroundCompaction() {
 
   mutex_.AssertHeld();
 
-#if 1
+#if 0
   if (imm_ != NULL) {
     pthread_rwlock_rdlock(&gThreadLock0);
     status=CompactMemTable();
@@ -937,7 +939,7 @@ Status DBImpl::BackgroundCompaction() {
   bool is_manual = (manual_compaction_ != NULL);
   InternalKey manual_end;
   if (is_manual) {
-    ManualCompaction* m = manual_compaction_;
+    ManualCompaction* m = (ManualCompaction *) manual_compaction_;
     c = versions_->CompactRange(m->level, m->begin, m->end);
     m->done = (c == NULL);
     if (c != NULL) {
@@ -1004,7 +1006,7 @@ Status DBImpl::BackgroundCompaction() {
   }
 
   if (is_manual) {
-    ManualCompaction* m = manual_compaction_;
+    ManualCompaction* m = (ManualCompaction *)manual_compaction_;
     if (!status.ok()) {
       m->done = true;
     }
@@ -1374,7 +1376,7 @@ struct IterState {
   port::Mutex* mu;
   Version* version;
   MemTable* mem;
-  MemTable* imm;
+  volatile MemTable* imm;
 };
 
 static void CleanupIteratorState(void* arg1, void* arg2) {
@@ -1399,7 +1401,7 @@ Iterator* DBImpl::NewInternalIterator(const ReadOptions& options,
   list.push_back(mem_->NewIterator());
   mem_->Ref();
   if (imm_ != NULL) {
-    list.push_back(imm_->NewIterator());
+     list.push_back(((MemTable *)imm_)->NewIterator());
     imm_->Ref();
   }
   versions_->current()->AddIterators(options, &list);
@@ -1447,7 +1449,7 @@ Status DBImpl::Get(const ReadOptions& options,
   }
 
   MemTable* mem = mem_;
-  MemTable* imm = imm_;
+  volatile MemTable* imm = imm_;
   Version* current = versions_->current();
   mem->Ref();
   if (imm != NULL) imm->Ref();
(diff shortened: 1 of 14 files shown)


kellymclaughlin:
I see a test failure running `make test` on eleveldb with `1.4.2-turner` checked out for eleveldb and the leveldb dependency; however, I see the same failure on `master` so I am assuming it is unrelated to these changes.

```
cacheleak: cacheleak_test_ (module 'cacheleak')...RSS1: 304908
RSS1: 353980
*failed*
in function cacheleak:'-cacheleak_loop/3-fun-2-'/2 (test/cacheleak.erl, line 66)
in call from cacheleak:cacheleak_loop/3 (test/cacheleak.erl, line 66)
**error:{assertion_failed,[{module,cacheleak},
                   {line,66},
                   {expression,"MaxFinalRSS > RSS"},
                   {expected,true},
                   {value,false}]}


module 'basho_bench_driver_eldb'
=======================================================
  Failed: 1.  Skipped: 0.  Passed: 16.
```

matthewvon:
cache leak test fails 3 out of 4 times ... minimum.  Jon has previously said that we should likely remove it.  But whenever I get ready to do that ... it seems to work again.

kellymclaughlin:
I did a Riak CS riak_test run using this branch for a functional sanity check and all tests passed.

kellymclaughlin:
:+1: on the PosixMmapFile changes

db/db_impl.cc
```diff
@@ -243,30 +246,36 @@ void DBImpl::MaybeIgnoreError(Status* s) const {
...
 
-      filenames.clear();
-      dirname=MakeDirName2(dbname_, level, "sst");
-      env_->GetChildren(dirname, &filenames); // Ignoring errors on purpose
+      // prune the database root directory
+      std::vector<std::string> filenames;
+      env_->GetChildren(dbname_, &filenames); // Ignoring errors on purpose
       for (size_t i = 0; i < filenames.size(); i++) {
-          KeepOrDelete(filenames[i], level, live);
+          KeepOrDelete(filenames[i], -1, live);
```
andrewjstone, line 263: What's up with the -1 here ?

matthewvon, line 263: The github differencing logic is piecing stuff from high and low.  The real change is that I inserted "if (RunningCompactionCount()<2)", then tabbed everything else in the function in by one tab.  Old (red) line 248 is now new (green) line 256.  Move 7 lines down in red, then in green ... both have same "KeepOrDelete(filenames[i], -1, live)".  Move 19 lines down from the starting point.  You will see "KeepOrDelete(filenames[i], level, live);" in both.

db/db_impl.cc
```diff
@@ -810,7 +819,7 @@ void DBImpl::MaybeScheduleCompaction() {
   // writing of memory to disk: high priority
   if (NULL!=imm_)
```
andrewjstone, line 820: this seems to be a clause with nothing in it anymore. Should it be removed?

matthewvon, line 820: You are correct on both points.  I left the dead code in place in case something goes wrong and I need to know quicker "how it used to work".  This intent applies to the extra logging messages added, but should go away once Turner is "happy".

db/db_impl.h
```diff
@@ -62,6 +62,10 @@ class DBImpl : public DB {
   // file at a level >= 1.
   int64_t TEST_MaxNextLevelOverlappingBytes();
 
+  void BackgroundImmCompactCall();
+  bool IsCompactionScheduled() {return(bg_compaction_scheduled_ || NULL!=imm_);};
+  uint32_t RunningCompactionCount() {return(running_compactions_);};
```
andrewjstone, line 67: running_compactions_ is accessed within a mutex elsewhere. not sure if this matters

matthewvon, line 67: You are absolutely correct.  The value is changed within a mutex (otherwise it would have to be changed via atomic operations).  Here it is only accessed.  The variable is marked as "volatile" so the compiler will not optimize the retrieval.  And I believe the function is only used today in DeleteObsoleteFiles() which is called when the same mutex is held.  But even if not, volatile gives me the level of thread security I desire.

matthewvon, line 67: I will add a comment concerning the mutex assumptions when this moves to a real release.

andrewjstone, line 67: That makes sense. Thanks for the explanation.

jtuple:
FYI, Jon has asked me to look over this PR as well. Just finished reading description in wiki, digging into the code now. Since Kelly/Andrew are already reviewing, I'm going to try to work though things quickly. Jon simply wanted someone who was familiar with the async eleveldb threading changes (eg. me) to be in the loop.

andrewjstone:
Thanks @jtuple.

After looking over this code I don't see any obvious errors. Being that my c++ is rusty and I don't quite understand all the logic here I'm going to give this a 'not -1' h/t @kellymclaughlin

db/db_impl.cc
```diff
@@ -863,7 +872,10 @@ void DBImpl::BGWork(void* db) {
 
 void DBImpl::BackgroundCall() {
   MutexLock l(&mutex_);
-  assert(bg_compaction_scheduled_);
+  ++running_compactions_;
+  Log(options_.info_log, "Background compact start: %u", running_compactions_);
+
+  assert(IsCompactionScheduled());
```
jtuple, line 878: Minor note: this assertion should probably be changed to only check for `bg_compaction_scheduled_` as the other condition that makes `IsCompactScheduled` true is a non-NULL `imm`, but `imm` compaction isn't handled by `BackgroundCall` anymore.

db/db_impl.cc
```diff
@@ -880,31 +892,81 @@ void DBImpl::BackgroundCall() {
...
+      // the problem.
+      bg_cv_.SignalAll();  // In case a waiter can proceed despite the error
+      Log(options_.info_log, "Waiting after background imm compaction error: %s",
+          s.ToString().c_str());
+      mutex_.Unlock();
+      env_->SleepForMicroseconds(1000000);
+      mutex_.Lock();
+    }
+  }
+//  IsCompactionScheduled() = false;
```
jtuple, line 928: Note sure if this is commented out old code, or a comment. If a comment, I believe it is incorrect -- we could still have a running concurrent compaction (eg. `bg_compaction_scheduled_`).

db/memtable.h
```diff
@@ -24,10 +24,10 @@ class MemTable {
   explicit MemTable(const InternalKeyComparator& comparator);
 
   // Increase reference count.
-  void Ref() { ++refs_; }
+  void Ref() volatile { ++refs_; }
 
   // Drop reference count.  Delete if no more references exist.
-  void Unref() {
+  void Unref() volatile {
```
jtuple, line 30: Here in `memtable.h`, you have references that use non-volatile variables but volatile accessors functions. In most other places, such as `atomics.h` and the `ref_count`/`ReleaseRef` code in `env_posix.cc` you have volatile variables with non-volatile accessor functions. What's the difference between these two usage scenarios?

db/version_set.cc
```diff
@@ -1469,6 +1484,28 @@ void VersionSet::SetupOtherInputs(Compaction* c) {
                                          &c->grandparents_);
       }
   }   // if
+#if 1
+  // compacting into an overlapped layer
```
jtuple, line 1488: I don't believe this change (lines 1487 to 1508) was discussed in the wiki summary. Is this just an additional performance optimization?

util/hot_threads.cc
```diff
@@ -0,0 +1,284 @@
...
+        // a work item identified (direct or queue), work it!
+        //  then loop to test queue again
+        if (NULL!=submission)
+        {
+           (*submission)();
+//    basho::async_nif::work_result result = work_item();
+//            HotThreadPool::notify_caller(*submission);
+
+            // resubmit will increment reference again, so
+            //  always dec even in reuse case
```
jtuple, line 94: I know you'll handle this when you get around to cleanup, but figured it doesn't hurt to have a note here. Seems like the above commented out code as well as the comment about resubmitting/recycling doesn't apply to the LevelDB version of the thread pool and can be deleted.

jtuple:
Compared `atomics.h` against equivalent code in eleveldb. All looks good. I notice that there's two new functions in `atomics.h` that weren't in eleveldb: `add_and_fetch` and `sub_and_fetch`. Neither are called anywhere inside LevelDB currently, so I cannot claim to have tested them. But, they look reasonable on visual inspection.

jtuple:
When comparing `hot_threads.cc` against eleveldb's thread pool, I noticed that we have a race condition in both that could lead to a deadlock. It's basically the issue we tried to address in basho/eleveldb#50

Consider the code in [hot_threads.cc#L108-L114](https://github.com/basho/leveldb/blob/8b5c6bc178bc95339d0d1b64aa7f83e7af6d1f20/util/hot_threads.cc#L108-L114):

``` c++
            // only wait if we are really sure no work pending
            if (0==m_Pool.m_WorkQueueAtomic)
            {
                // yes, thread going to wait. set available now.
                m_Available=1;
                m_Condition.Wait();
            }    // if
```

As well as the code in [hot_threads.cc#L260-L268](https://github.com/basho/leveldb/blob/8b5c6bc178bc95339d0d1b64aa7f83e7af6d1f20/util/hot_threads.cc#L260-L268):

``` c++
            // no waiting threads, put on backlog queue
            {
                SpinLock lock(&m_QueueLock);
                inc_and_fetch(&m_WorkQueueAtomic);
                m_WorkQueue.push_back(item);
            }

            // to address race condition, thread might be waiting now
            FindWaitingThread(NULL);
```

If the worker thread is preempted after performing the conditional check, but before setting `m_Available` to 1 you can end up with an interleaving such as:

``` c++
/* thread 1: */   if (0==m_Pool.m_WorkQueueAtomic)
/* thread 2: */   inc_and_fetch(&m_WorkQueueAtomic);
/* thread 2: */   m_WorkQueue.push_back(item);
/* thread 2: */   FindWaitingThread(NULL);
/* thread 1: */   m_Available=1;
/* thread 1: */   m_Condition.Wait();
```

Thus, thread 2 queues work, tries to find an available thread, fails to find thread 1 and returns. Afterwards, thread 1 goes to sleep, never to be signaled.

I'm not sure how likely this is to ever occur. But, I was able to reliably hit this condition by using a 1-thread pool and adding a small busywait wait before the `m_Available=1` line. Relevant commit on `jdb-1.4.2-turner` branch [here](https://github.com/basho/leveldb/commit/de00b2a5d34ae2c5b2ce206004f92e5abde05674).

To make things clear, I added additional logging in [this](https://github.com/basho/leveldb/commit/29a4148b6885bf46b8d3ff25c06e619b929ec758) commit, where thread events are logged to `/tmp/thread.log`. 

A combined log of thread + LevelDB events for a deadlock scenario is [here](https://gist.github.com/jtuple/2231168aa8849caa6fcf). As shown, we get into a state where we hit `waiting 2` and stall forever, because the enqueued background imm task never runs. All LevelDB traffic stops, throughput goes to zero, etc.

All in all, I have no idea how likely this race is to matter in the real world, with >1 background thread and no artificial delay. But, think it is still a remote possibility. Thoughts?

db/db_impl.cc
matthewvon, line 878: no.  Today, bg_compaction_scheduled_ is a flag for any compaction BUT imm compaction.  This routine should only be called for a non-imm compaction AND it should only be called when bg_compaction_scheduled_ is set.

db/db_impl.cc
matthewvon, line 928: This is one of the places where "dead code" should be cleared.  I left the dead line as a reminder of the sequence used in the original function.  Honestly it took two, maybe three, passes at this function to get it proper for simultaneous compactions ... with this being the imm compaction.  The dead code is still a reminder if we find another issue at customer site next week.  Will be cleared by the time this goes to 2.0.  The imm compaction's "flag" is whether or not DBImpl::imm_ is set or not.

db/memtable.h
matthewvon, line 30: Damn good question.  This was in response to a compiler warning.  I made the DBImpl::imm_ variable a volatile as a precaution against compiler optimizations.  imm_ is a MemTable object.  When I call imm_->Unref(), there is a compiler warning about loss of volatility if Unref() is also not volatile.  The issue is that a volatile object must have volatile member functions (so says the compiler).  Atomics and ref_count operate on atomic variables but they are either static functions or variables/operations within non-volatile objects.

db/version_set.cc
matthewvon, line 1488: First, this change was made before the UpdateStats() function was modified.  UpdateStats() change fixed the main problem with "premature compactions".  However, there is still the possibility of an overlapped level (name 0 and 1 today) having 4 or more files will only compacting a subset of the files.  I changed a ton of code in 1.4.0 on the assumption that overlapped levels always compacted all files.

So this is overkill for the issue that needed to be addressed.  I am comfortable removing from this release if it concerns you.

util/hot_threads.cc
matthewvon, line 94: You are correct.  Failure to tidy up on my part.

matthewvon:
Joe comment #1:  the two "new" functions were added as part of mv-flexcache branch of Riak 2.0.  I pulled the file from there.  But you are correct that you might be the first person reviewing them.

Joe comment #2:  I agree with your assessment.  The chance of the stall in eleveldb today is very slim since there are 71 threads (a prime number larger than 64 which is a common vnode count).  The thread counts will be much lower in leveldb's usage.  This is a bug that needs to be addressed.  I am comfortable with it going to this targeted, high volume customer.  But low traffic sites are susceptible.

db/db_impl.cc
jtuple, line 878: I agree with you. This assertion isn't checking that case. It's checking `IsCompactionScheduled()` which is true when `bg_compaction_scheduled_` OR `imm != NULL`. If we somehow end up in this function when `bg_compaction_scheduled = false` but `imm != NULL`, the assertion should fire. Obviously, this doesn't affect things in production, but seems helpful to have proper assertion to help with future debugging in later code changes.

jtuple:
Spent a lot of time pushing this branch both inside and outside of Riak. In my non-Riak testing, I used a micro-benchmarck that tested having 10 to 128 concurrent open DBs with 10 to 400 concurrent Erlang processes per DB issuing reads/writes. This is much more concurrency than Riak normally can provide. Multiple runs over 2 hours along with different object sizes, read/write ratios, etc all worked without a hitch. Wasn't testing performance -- was trying to hit any race conditions or other concurrency issues.

This runs alternated from starting with fresh DBs on each run, to reusing a data created on a prior run. Likewise, I also 'kill -9' the Erlang VM on several of these runs, such that the follow-up run was starting from data that needed to run recovery. Things appeared to work as intended. Looking at the LOGs shows recovery kicking in when needed on startup, shows both background imm writes and compactions occurring as necessary, etc.

I've carefully looked over the new `hot_threads` code, the new mmap ref counting logic in `env_posix`, and the `atomics` changes, and other than the race condition noted above, everything else looks good. I also spent some time sketching out the old vs new approach to flushing imm tables, as well as manually traversed all the call paths, and it's clear to me that all paths that could flush imm tables previous cannot occur in the code as written today and only the new background imm flusher will do the work -- exactly as intended.

Once this is stabilized, I do think code cleanup is necessary not only for "cleanup" purposes, but also to prevent us accidentally reintroducing code that may concurrently call `CompactMemTable` when it is no longer safe (eg. accidentally calling `PrioritizeWork` in a future change).

Other than the noted race condition, there is nothing else I'm worried about in this code at this time. I agree that >1 thread pool is unlikely to hit the case I noted. The `hot_threads` version of the deadlock is also easy to identify: waiting 2 in the LOG that goes on forever, so easy to keep on eye out for. The related potential deadlock in the eleveldb thread pool is harder to identify, but we have been running that code in production for 9 months now, so no reason to hold things up in this case.

+1 for customer deployment with this branch. Agree with @matthewvon about fixing race condition for 2.0.

db/db_impl.cc
matthewvon, line 878: argh.  I misread everything.  assertion needs to change.

matthewvon:
issue addressed in mv-hot-threads2

PR #98, opened 2013-10-13
Mv counters update

matthewvon:
This branch updates the performance counters Init() function to properly resize the shared memory as variables get added to the code.  Previously, a user would have to manually "ipcrm" the shared memory segment (or reboot the computer) to have counters work after leveldb upgrade.

perf_count_test was created to validate the changes.  It is now part of the standard "make check" tests.

### DIFF
--- a/util/perf_count.cc
+++ b/util/perf_count.cc
@@ -1,8 +1,8 @@
 // -------------------------------------------------------------------
 //
-// perf_count.cc:  performance counters LevelDB (http://code.google.com/p/leveldb/)
+// perf_count.cc:  performance counters LevelDB
 //
-// Copyright (c) 2012 Basho Technologies, Inc. All Rights Reserved.
+// Copyright (c) 2012-2013 Basho Technologies, Inc. All Rights Reserved.
 //
 // This file is provided to you under the Apache License,
 // Version 2.0 (the "License"); you may not use this file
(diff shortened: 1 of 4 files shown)


util/perf_count.cc
```diff
@@ -205,23 +206,63 @@
...
+        // does the shared memory already exists (and of proper size if writing)
+        should_create=(0!=ret_val || ((shm_info.shm_segsz < sizeof(PerformanceCounters)) && !IsReadOnly));
+
+        // should old shared memory be deleted?
+        if (should_create && 0==ret_val)
+        {
+            ret_val=shmctl(id, IPC_RMID, &shm_info);
+            if (0!=ret_val)
+            {
+                syslog(LOG_ERR, "shmctl IPC_RMID failed [%d, %m]", errno);
```
jonmeredith, line 239: Never used %m before, should this be 

```
syslog(LOG_ERR, "shmctl IPC_RMID failed [%d, %m]", errno, errno);
```

Or is errno read from TLS?

matthewvon, line 239: (from http://linux.die.net/man/3/printf)

m

(Glibc extension.) Print output of strerror(errno). No argument is required.

jonmeredith, line 239: Happily works on Linux/OSX too as written.

jonmeredith:
+1 merge.

PR #100, opened 2013-10-16
Mv flexcache2

matthewvon:
Make flexcache default actions more friendly to developer builds of Riak.

### DIFF
--- a/util/env_posix.cc
+++ b/util/env_posix.cc
@@ -48,6 +48,8 @@ static Status IOError(const std::string& context, int err_number) {
   return Status::IOError(context, strerror(err_number));
 }
 
+volatile size_t gMapSize=20*1024*1024L;
+
 // background routines to close and/or unmap files
 static void BGFileCloser(void* file_info);
 static void BGFileCloser2(void* file_info);
@@ -267,10 +269,6 @@ class PosixMmapFile : public WritableFile {
       last_sync_ = NULL;
       dst_ = NULL;
 
-      // Increase the amount we map the next time, but capped at 1MB
-      if (map_size_ < (1<<20)) {
-        map_size_ *= 2;
-      }
     }
     else if (and_close && -1!=fd_)
     {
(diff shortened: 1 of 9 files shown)


util/env_posix.cc
```diff
@@ -267,10 +269,6 @@ class PosixMmapFile : public WritableFile {
       last_sync_ = NULL;
       dst_ = NULL;
 
-      // Increase the amount we map the next time, but capped at 1MB
```
engelsanchez, line 270: Could you explain what part this is playing here?

matthewvon, line 270: This code has been dead since the day we raised the map_size_ from 4096 bytes to 20Mbytes.  And now we are further "managing" the value through the (cough, cough) global gMapSize.  It is time to just remove the code.

engelsanchez, line 270: Ok, sounds good.

joedevivo:
+1 I am now able to start riak-ee on my laptop with 8gb of ram.

util/flexcache.cc
```diff
@@ -77,24 +75,22 @@
...
 
     // what is total memory assigned to a type
-    if (IsInternal)
+    if (IsInternal && 0!=user_count)
         shared_total=(m_TotalMemory*2)/10;  // integer *.2
-    else if (0!=internal_count)
+    else if (!IsInternal && 0!=internal_count)
         shared_total=(m_TotalMemory*8)/10;
-    else // no internal database
+    else // no databases in "other" category
```
engelsanchez, line 89: So now in a node with only internal databases they are free to use all of the total memory for themselves. So we have to worry about setting total memory available properly in a bitcask node. That is, you don't want these internal databases to steal memory from bitcask.

engelsanchez:
Hmmm... @matthewvon did you see my note about the change to internal databases being able to allocate the 100% of the total memory? I believe this is a problem.

util/flexcache.cc
matthewvon, line 89: agreed ... all points

matthewvon:
ok, does the kv/core committee have a recommended default setting?

PR #101, opened 2013-10-16
Mv hot threads1

matthewvon:
This is branch contains the threading code originally developed for the eleveldb async NIF.  It was previously delivered in leveldb as part of a patch in branch 1.2.4-turner.  Joe B. spotted an unlikely, but real, race condition that this branch addressed via the QueueThread object.

Today only Imm to Level-0 compactions post to a hot thread pool.  The goal is to move background disk writes and all background compactions to a series of specialized hot thread pools ... but that will be a different branch.

### DIFF
--- a/util/env_posix.cc
+++ b/util/env_posix.cc
@@ -28,9 +28,11 @@
 #include "port/port.h"
 #include "util/crc32c.h"
 #include "util/db_list.h"
+#include "util/hot_threads.h"
 #include "util/logging.h"
 #include "util/mutexlock.h"
 #include "util/posix_logger.h"
+#include "util/thread_tasks.h"
 #include "util/throttle.h"
 #include "db/dbformat.h"
 #include "leveldb/perf_count.h"
@@ -1129,6 +1131,7 @@ pthread_t PosixEnv::StartThread(void (*function)(void* arg), void* arg) {
   return(t);
 }
 
+
 // this was a reference file:  unmap, purge page cache, close
 void BGFileCloser(void * arg)
 {
@@ -1361,6 +1364,9 @@ static void InitDefaultEnv()
         crc32c::SwitchToHardwareCRC();
 
     PerformanceCounters::Init(false);
+
+    gImmThreads=new HotThreadPool(7, ePerfDebug1, ePerfDebug2,
+                                  ePerfDebug3, ePerfDebug4);
     started=true;
 }
 
@@ -1389,6 +1395,9 @@ void Env::Shutdown()
 
     ComparatorShutdown();
     DBListShutdown();
+
+    delete gImmThreads;
+    gImmThreads=NULL;
 
 }   // Env::Shutdown
 
--- a/util/env_posix.cc
+++ b/util/env_posix.cc
@@ -22,6 +22,7 @@
 #if defined(LEVELDB_PLATFORM_ANDROID)
 #include <sys/stat.h>
 #endif
+#include "leveldb/atomics.h"
 #include "leveldb/env.h"
 #include "leveldb/filter_policy.h"
 #include "leveldb/slice.h"
@@ -48,9 +49,6 @@ static Status IOError(const std::string& context, int err_number) {
 }
 
 // background routines to close and/or unmap files
-static void BGFileCloser(void* file_info);
-static void BGFileCloser2(void* file_info);
-// currently unused static void BGFileUnmapper(void* file_info);
 static void BGFileUnmapper2(void* file_info);
 
 // data needed by background routines for close/unmap
@@ -60,14 +58,17 @@ struct BGCloseInfo
     void * base_;
     size_t offset_;
     size_t length_;
-    size_t unused_;
+    volatile uint64_t * ref_count_;
     uint64_t metadata_;
 
     BGCloseInfo(int fd, void * base, size_t offset, size_t length,
-                size_t unused, uint64_t metadata)
+                volatile uint64_t * ref_count, uint64_t metadata)
         : fd_(fd), base_(base), offset_(offset), length_(length),
-          unused_(unused), metadata_(metadata)
-    {};
+          ref_count_(ref_count), metadata_(metadata)
+    {
+        if (NULL!=ref_count_)
+            inc_and_fetch(ref_count_);
+    };
 };
 
 class PosixSequentialFile: public SequentialFile {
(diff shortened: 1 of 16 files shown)


jtuple:
I'm reviewing this code now. Also, since most of this was previously reviewed in #97, I went spent some time reconstructing a proper change base to review this branch against to make it easier to see what's new since the last review. So, reviewing based off [this changeset](https://github.com/basho/leveldb/compare/jdb-mv-hot-threads-base...jdb-mv-hot-threads-new).

util/env_posix.cc
```diff
@@ -237,32 +215,25 @@ class PosixMmapFile : public WritableFile {
...
-      if (!is_write_only_)
+      if (!is_async_)
       {
-          if (and_close)
-              BGFileCloser2(ptr);
-          else
-              BGFileUnmapper2(ptr);
+          BGCloseInfo * ptr=new BGCloseInfo(fd_, base_, file_offset_, limit_-base_,
+                                            NULL, metadata_offset_);
+          BGFileUnmapper2(ptr);
```
jtuple, line 230: In this case (the non-async close case), we don't ever delete the allocated `BGCloseInfo` because [line 1209](https://github.com/basho/leveldb/blob/1b3b6ab35eb794e51eefee131c2ac04542c489fd/util/env_posix.cc#L1209) was changed to assume that `BGCloseInfo` was submitted to/deallocated by the thread pool -- which isn't true in this case. I think we just need a `delete ptr` after the `BGFileUnmapper2(ptr)` call here.

util/hot_threads.cc
```diff
@@ -0,0 +1,369 @@
...
+                // retest with locking
+                SpinLock lock(&m_Pool.m_QueueLock);
+
+                if (!m_Pool.m_WorkQueue.empty())
+                {
+                    submission=m_Pool.m_WorkQueue.front();
+                    m_Pool.m_WorkQueue.pop_front();
+                    dec_and_fetch(&m_Pool.m_WorkQueueAtomic);
+                    m_Pool.IncWorkDequeued();
+                    m_Pool.IncWorkWeighted(2);
```
jtuple: Do we want hard coded `IncWorkWeighted(2)` here or do we want to switch to the approach you used in [lines 171-172](https://github.com/basho/leveldb/blob/1b3b6ab35eb794e51eefee131c2ac04542c489fd/util/hot_threads.cc#L171-L172):

``` c++
m_Pool.IncWorkWeighted(Env::Default()->NowMicros()
                       - submission->m_QueueStart);
```

util/hot_threads.cc
```diff
@@ -0,0 +1,369 @@
...
+
+}   // QueueThreadStaticEntry
+
+
+QueueThread::QueueThread(
+    class HotThreadPool & Pool)
+    : m_Pool(Pool)
+{
+    memset(&m_Semaphore, 0, sizeof(m_Semaphore));
+    sem_init(&m_Semaphore, 0, 0);
```
jtuple: So, it turns out that OS X doesn't support `sem_init`, only named semaphores created via `sem_open`. This call fails (and we don't check the error condition here), leading to the future use of `sem_wait` on [line 190](https://github.com/basho/leveldb/blob/1b3b6ab35eb794e51eefee131c2ac04542c489fd/util/hot_threads.cc#L189-L190) to immediately return with an error (which we also don't check for), causing the safety thread to loop endlessly. On OS X, opening a LevelDB instance without even reading/writing to it results in 200% CPU usage while the two safety threads busy wait.

On Linux, everything works as intended.

I imagine this is something we need to fix. While no one uses OS X in production, it is a common Riak development platform and increasingly the first platform someone who downloads Riak first tries it on. For better or worse.

Should we just fall back to using named semaphores inside the relevant LevelDB directory via `sem_open` when on OS X?

util/hot_threads.cc
```diff
@@ -0,0 +1,369 @@
...
+                SpinLock lock(&m_QueueLock);
+                inc_and_fetch(&m_WorkQueueAtomic);
+                m_WorkQueue.push_back(item);
+            }
+
+            // to address race condition, thread might be waiting now
+            FindWaitingThread(NULL);
+
+            // to address second race condition, send in QueueThread
+            sem_post(&m_QueueThread.m_Semaphore);
```
jtuple: Is it worthwhile to optimize things and only do the `sem_post` if the `FindWaitingThread(NULL)` call returns false?

jtuple:
Ran branch through the same stress testing I did in #97. As noted above, OS X had 200% CPU usage from the safety threads that were never sleeping, but worked fine otherwise. Linux was fine across the board.

I reintroduced the changes that help trigger the known deadlock. With the `sem_post` in submit commented out, I could hit the deadlock as expected. With `sem_post` uncommented, things didn't deadlock. Adding print statements showed the safety thread picking up the slack as necessary.

All looks good other than the issues I mentioned in the diff comments above.

util/hot_threads.cc
matthewvon: I coded that optimization the first time, then removed it.  My thought was that this is my safety thread and should just launch it in case of a race we have not found.  I can be easily convinced to put that back.

util/hot_threads.cc
matthewvon: Will fix.  Classic case of why copy/paste of highly similar code leaves one wrong when bug (code update) happens.

util/env_posix.cc
matthewvon, line 230: Good find.  I have not rerun valgrind against recent code.  Will add refine and refdec pair so that same logic works in both conditions.  Refdec to zero deletes object.

util/hot_threads.cc
matthewvon: Now I remember ... FindWaitingThread(NULL) is outside any type of lock/synchronization object.  I therefore assume that a race condition exists between its scan and the proposed "if" block.  Instead of attempting to prove / disprove the race condition, I took the safe path of always calling sem_post.  This code is already overdue and the optimization is not worth it.

util/hot_threads.cc
matthewvon: I am not against named semaphores.  I am against trying to work out a naming scheme that will work in a developer environment ... need unique names by pool by Riak instance.  I am too lazy to do that today since Jared will be building soon and I fear temp names are not going to be trivial.  QueueThread is now only operational if sem_init() succeeds.

I will happily accept an "issue" to correct before final release.

jtuple:
Looked over changes and retested on OS X and Linux. We should fix the semaphore creation issue before 2.0 final, but no reason to hold up merging now. +1 merge.

PR #102, opened 2013-10-16
revert to original, fixed 20 percent for internal databasses

matthewvon:


engelsanchez:
:+1: Riak can start just fine and memory assignment to instances both internal and vnodes looks sane with leveldb as a backend and without.  Sorry for the delay in the review. I hit some weird problems that seem to have gone away with a clean build.

PR #103, opened 2013-10-21
install additional performance counters

matthewvon:
add counters that have been missing.  give each new hot thread pool counters (instead of the debug counters)


coderoshi:
My C is not the strongest, but this does indeed look to be just adding counters.

coderoshi:
:+1: for clarity

PR #104, opened 2013-10-23
change build_detect_platform to check for OS X Mavericks

vinoski:
On Mavericks clang is the default compiler, so modify build_detect_platform
to detect that OS X version and set the necessary compiler and linker
switches.

Successfully tested on Mavericks, Mountain Lion, and Lion.

I strongly suggest rebasing this to master rather than merging, as that way the corresponding changes in the eleveldb c_src/build_deps.sh script will continue to pick up the right leveldb commit sha created by this commit.

### DIFF
--- a/build_detect_platform
+++ b/build_detect_platform
@@ -59,7 +59,16 @@ PLATFORM_SHARED_VERSIONED=true
 case "$TARGET_OS" in
     Darwin)
         PLATFORM=OS_MACOSX
-        COMMON_FLAGS="-fno-builtin-memcmp -DOS_MACOSX"
+        oIFS="$IFS"; IFS=.
+        set `uname -r`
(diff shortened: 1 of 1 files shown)


matthewvon:
Discussed offline with vinoski that the -fno-builtin-memcmp option was added to leveldb by Google to address a severe performance difference between compiler generated memcmp and glibc's memcmp.  Suggestion is to utilize clang's -fno-builtin option which is supposed to perform the same trade of functions.  vinoski is retesting.

vinoski:
Based on the db_check program, it appears the -fno-builtin option slows things down slightly as compared to not using it, so leaving it out.

matthewvon:
verified this does not break build of osx 10.8 and Debian linux.  seems everyone has had favorable experience.  
+1

PR #105, opened 2013-10-31
Mv flexcache5

matthewvon:
Clean up of the bytes used accounting in flexcache.  Both the block cache and the file cache objects were missing byte counts of "attached" objects.  These missing byte counts caused flexcache to under-account for bytes in use.  Therefore the total amount of memory used by the caches was actually greater than the cache limits.


vinoski:
Looks good to merge, +1. Eyeballed all the diffs, compiled and ran all tests successfully.

PR #108, opened 2013-11-24
Mv aggressive delete

matthewvon:
branch discussion is here:  https://github.com/basho/leveldb/wiki/Mv-aggressive-delete


andrewjstone:
:+1:

PR #110, opened 2013-12-04
mv-fadvise-control

matthewvon:
Changes discussed here https://github.com/basho/leveldb/wiki/mv-fadvise-control


engelsanchez:
+1  I've reviewed the code and tested under Linux using strace to see actual fadvise usage (in Linux it uses fadvise64). Works as advertised.

This PR needs to be retargeted to point to 1.4 though.

matthewvon:
This was merged manually to the 1.4 branch.

PR #111, opened 2013-12-05
Mv dynamic block size

matthewvon:
Branch description is here:  https://github.com/basho/leveldb/wiki/mv-dynamic-block-size


db/db_impl.cc
```diff
@@ -1087,11 +1093,163 @@ Status DBImpl::OpenCompactionOutputFile(CompactionState* compact) {
...
+
+                  // did size change?
+                  if (options.block_size!=old_size)
+                  {
+                      gPerfCounters->Inc(ePerfDebug0);
+                      block_size_changed_=env_->NowMicros();
+                  }   // if
+              }   // if
+
+              last_low_mem_=env_->NowMicros();
```
andrewjstone: Maybe set `last_low_mem_=block_size_changed_` instead of calling `env->NowMicros()` again?

db/db_impl.cc
```diff
@@ -1087,11 +1093,163 @@ Status DBImpl::OpenCompactionOutputFile(CompactionState* compact) {
...
+
+          now=env_->NowMicros();
+
+          if (!double_cache.GetPlentySpace())
+          {
+              // do not make changes often, a multi file compaction
+              //  could raise more than one step (5 min)
+              if (block_size_changed_+(5*60*1000000L) < now)
+              {
+                  size_t old_size=current_block_size_;
```
andrewjstone, line 1118: The temporary variable `old_size` can be removed if `MaybeRaiseBlockSize` doesn't mutate `current_block_size_`. The mutation can then happen inside the modified `if (options.block_size !=current_block_size_)` block. It also prevents an extra conditional in MaybeRaiseBlockSize.

db/db_impl.cc
```diff
@@ -1087,11 +1093,163 @@ Status DBImpl::OpenCompactionOutputFile(CompactionState* compact) {
...
+    {
+        size_t high_size, low_size, cur_size, increment, file_data_size, file_key_size;
+
+        // 2a. Highest block size:
+        //      (sqrt()/sqrt() stuff is from first derivative to minimize
+        //       total read size of one block plus file metadata)
+
+        // limited by keys or filesize? (pretend metadata is zero, i love pretend games)
+        file_data_size=versions_->MaxFileSizeForLevel(CompactionStuff.level());
+        file_key_size=file_data_size / avg_value_size;
```
andrewjstone: `file_key_size` is really a count and not a size. I think something like `keys_per_file` would be more descriptive.

db/db_impl.cc
```diff
@@ -1087,11 +1093,163 @@ Status DBImpl::OpenCompactionOutputFile(CompactionState* compact) {
...
+      }   // if
+      compact->builder = new TableBuilder(options, compact->outfile);
+  }   // if
+
   return s;
 }
 
+
+size_t
+DBImpl::MaybeRaiseBlockSize(
```
andrewjstone, line 1149: It'd be nice to have tests showing what block sizes we get from running this function against various values for the inputs, to ensure we don't hit any pathological cases. Graphs would help. Alternatively, or in conjunction with this, graphs of block size changes vs disk activity during performance tests would probably be informative. I'm sure you've thought of this and have all sorts of tuning tests planned, just adding some emphasis.

db/version_set.cc
```diff
@@ -1828,4 +1839,163 @@ void Compaction::ReleaseInputs() {
   }
 }
 
-}  // namespace leveldb
+/**
+ * Riak specific:  populate statics data about this compaction
```
andrewjstone: typo: statics -> statistics

util/cache2.cc
```diff
@@ -365,12 +365,25 @@ class ShardedLRUCache2 : public Cache {
...
 
       return;
 
   } // ShardedLRUCache2::Resize
 
+
+  // let doublecache know state of cache space
+  void SetFreeSpaceWarning(size_t FileMetaSize)
+  {
+      bool plenty_space;
```
andrewjstone, line 379: I love the name of this variable. That is all.

include/leveldb/options.h
```diff
@@ -119,6 +119,15 @@ struct Options {
   // Default: 4K
   size_t block_size;
 
+  // Riak specific:  non-zero value activates code to automatically
+  // increase block_size as needed to ensure maximum number of files
+  // are available in the file cache.  The value indicates how many
+  // incremental increases to use between the original block_size
+  // and largest, reasonable block_size.
+  //
+  // Default: 4
```
andrewjstone: It defaults to 16 in `util/options.cc`

db/db_impl.cc
```diff
@@ -1087,11 +1093,163 @@ Status DBImpl::OpenCompactionOutputFile(CompactionState* compact) {
...
+    //
+    // 2. Define boundaries of block size steps.  Calculate
+    //    "next step"
+    //
+    if (0!=tot_user_data && 0!=tot_index_keys && 0!=avg_value_size
+        && 0!=avg_key_size && 0!=avg_block_size)
+    {
+        size_t high_size, low_size, cur_size, increment, file_data_size, file_key_size;
+
+        // 2a. Highest block size:
```
andrewjstone, line 1187: Should probably break this out into a method. I think there should be a method for calculating high_size and low_size, as well as a sanity check function to better split this code up, instead of having it all in one large method.

db/version_set.cc
```diff
@@ -1828,4 +1839,163 @@ void Compaction::ReleaseInputs() {
...
+                temp_cnt=0;
+                user_est=0;
+                idx_est=0;
+
+                // get and hold handle to cache entry
+                s=tables.TEST_FindTable(fmd->number, fmd->file_size, fmd->level, &handle);
+
+                if (s.ok())
+                {
+                    // 1. total size of all blocks before compaction
```
andrewjstone, line 1885: I think each numbered comment should be it's own function. It would remove the need for the comments :) It'd also shorten this function. If performance is a concern, they can be inlined.

db/db_impl.cc
matthewvon: The intended change was to have all env_->NowMicros() calls use variable "now".  Must have gotten distracted.  Fixing that and moving last_low_mem_ higher in the block with a descriptive comment.

db/db_impl.cc
matthewvon, line 1118: I declared current_block_size_ as a volatile.  This is because Riak 2.0 allows simultaneous compactions on independent levels.  I believe this function is already protected by the database level mutex … but to lazy to guarantee that today, and not willing to risk it in the future … cuz I have burned myself once already on "knowing how locks worked" in the code, then changed a code path breaking the assumption elsewhere.  These values are not critical, so I do not believe two threads working through here at the same time will interact horribly or leave values "wrong" … hence why I am intentionally not thread protecting actively.

I should mention that this lazy threading model is to avoid creating yet another mutex or similar object.  Any mutex that is actually used by two threads simultaneously causes a disproportional performance loss related to Erlang schedulers.  Laziness pays off greatly in the shadow of Erlang.

db/db_impl.cc
matthewvon: changed as suggested.

db/version_set.cc
matthewvon: corrected

include/leveldb/options.h
matthewvon: good eye.  four was not a good default, steps were bigger than needed … suboptimal.

db/db_impl.cc
matthewvon, line 1187: hmm.  That is a stylistic thing.  Based upon how this was developed and tuned, I prefer to keep it this way.

db/version_set.cc
matthewvon, line 1885: I am more concerned about long term maintenance.  This function is truly a train of thought.  I don't want to break it up.  The estimation functions have a bit of overlap and breaking them up might lose sight of that fact when next developer has to do maintenance.

db/db_impl.cc
matthewvon, line 1149: Sent graphic image of max block size plane to Andrew offline.  No holes / spikes.  Similarly constructed spreadsheet with file sizes from 800 to 800M and key sizes from 10 to 1080 … resulting max block size matrix was a plane without spikes / holes.

db/db_impl.cc
andrewjstone, line 1118: Ok. Makes sense I think. If it gets changed underneath, you want to keep doing the calculation with the temporary.

db/db_impl.cc
andrewjstone, line 1149: Yep, things look good here.

db/version_set.cc
andrewjstone, line 1885: MvM prefrence. Ok by me.

andrewjstone:
:+1: Nice work @matthewvon

PR #112, opened 2013-12-09
move mv-async-close from 2.0 branch to 1.4.4

matthewvon:
Moving recent branch against 2.0 to the 1.4 tree.  Notes are here:  https://github.com/basho/leveldb/wiki/mv-async-close

### DIFF
--- a/util/env_posix.cc
+++ b/util/env_posix.cc
@@ -22,6 +22,7 @@
 #if defined(LEVELDB_PLATFORM_ANDROID)
 #include <sys/stat.h>
 #endif
+#include "leveldb/atomics.h"
 #include "leveldb/env.h"
 #include "leveldb/filter_policy.h"
 #include "leveldb/slice.h"
@@ -48,9 +49,6 @@ static Status IOError(const std::string& context, int err_number) {
 }
 
 // background routines to close and/or unmap files
-static void BGFileCloser(void* file_info);
-static void BGFileCloser2(void* file_info);
-// currently unused static void BGFileUnmapper(void* file_info);
 static void BGFileUnmapper2(void* file_info);
 
 // data needed by background routines for close/unmap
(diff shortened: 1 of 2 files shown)


util/env_posix.cc
```diff
@@ -356,12 +317,36 @@ class PosixMmapFile : public WritableFile {
...
+    // hard code
+    if (!is_async_)
+    {
+        int ret_val;
+
+        ret_val=ftruncate(fd_, file_length);
+        if (0!=ret_val)
+            syslog(LOG_ERR,"Close ftruncate failed [%d, %m]", errno);
+
+        ret_val=close(fd_);
```
bookshelfdave, line 338: you don't use the ret_val from `close`

util/env_posix.cc
```diff
@@ -400,6 +385,30 @@ class PosixMmapFile : public WritableFile {
...
+          int ret_val;
+
+          ret_val=dec_and_fetch(Count);
+          if (0==ret_val)
+          {
+              ret_val=ftruncate(File, *(Count +1));
+              if (0!=ret_val)
+                  syslog(LOG_ERR,"ReleaseRef ftruncate failed [%d, %m]", errno);
+
+              ret_val=close(File);
```
bookshelfdave, line 405: same thing, you don't use the ret_val from `close`

util/env_posix.cc
```diff
@@ -652,6 +662,9 @@ class PosixEnv : public Env {
     }
     gFileLocks.Remove(my_lock->name_);
     close(my_lock->fd_);
+
+    my_lock->fd_=-1;
```
bookshelfdave, line 666: setting the fd to -1 before checking for `close` errors

util/env_posix.cc
matthewvon, line 338: There used to be a syslog about the return value back when I was looking for the failure.  Removed the syslog() call but left the ret_val= as convenience … if I was wrong about having fixed the problem.

util/env_posix.cc
matthewvon, line 405: There used to be a syslog about the return value back when I was looking for the failure.  Removed the syslog() call but left the ret_val= as convenience … if I was wrong about having fixed the problem.

util/env_posix.cc
matthewvon, line 666: Was more worried about double closes, not close errors.  None of the original google code verifies close errors … and dumping them into syslog is operationally useless.

bookshelfdave:
+1

bookshelfdave:
also, `make check` shows all tests passing

matthewvon:
This was manually merged to the 1.4 branch.

PR #115, opened 2014-01-09
Mv aae writeblock

matthewvon:
Restore AAE exemption to writeblock size accounting in write throttle, BUT leave the accounting for 2i writeblocks. Required back-porting of Riak 2.0 is_internal_db flag.


jaredmorrow:
Closed in favor of #116

PR #117, opened 2014-01-17
mv-tuning4

matthewvon:
Description here:  https://github.com/basho/leveldb/wiki/mv-tuning4

### DIFF
--- a/db/corruption_test.cc
+++ b/db/corruption_test.cc
@@ -323,7 +323,9 @@ TEST(CorruptionTest, CompactionInputErrorParanoid) {
   DBImpl* dbi = reinterpret_cast<DBImpl*>(db_);
 
   // Fill levels >= 1 so memtable compaction outputs to level 1
-  for (int level = 1; level < config::kNumLevels; level++) {
+  //  matthewv 1/10/14 - what does "levels" have to do with this,
+  //  switching to compaction trigger.
+  for (int level = 1; level < config::kL0_CompactionTrigger; level++) {
     dbi->Put(WriteOptions(), "", "begin");
     dbi->Put(WriteOptions(), "~", "end");
     dbi->TEST_CompactMemTable();
--- a/db/dbformat.h
+++ b/db/dbformat.h
@@ -25,7 +25,8 @@ static const int kNumLevels = 7;
 static const int kNumOverlapLevels = 2;
 
 // Level-0 compaction is started when we hit this many files.
-static const size_t kL0_CompactionTrigger = 4;
+// Google:  static const size_t kL0_CompactionTrigger = 4;
+static const size_t kL0_CompactionTrigger = 6;
 
 // Soft limit on number of level-0 files.  We slow down writes at this point.
 static const size_t kL0_SlowdownWritesTrigger = 8;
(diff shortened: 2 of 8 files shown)


db/version_set.cc
```diff
@@ -57,13 +57,13 @@
...
 // WARNING: m_OverlappedFiles flags need to match config::kNumOverlapFiles ... until unified
 {
-    {10485760,  262144000,  57671680,      209715200,                0,     300000000, true},
-    {10485760,   82914560,  57671680,      419430400,                0,     209715200, true},
-    {10485760,  314572800,  57671680,     1006632960,        200000000,     314572800, false},
-    {10485760,  419430400,  57671680,     4094304000ULL,    3355443200ULL,  419430400, false},
-    {10485760,  524288000,  57671680,    41943040000ULL,   33554432000ULL,  524288000, false},
-    {10485760,  629145600,  57671680,   419430400000ULL,  335544320000ULL,  629145600, false},
-    {10485760,  734003200,  57671680,  4194304000000ULL, 3355443200000ULL,  734003200, false}
+    {10485760,  262144000,  57671680,      209715200,                 0,     420000000, true},
```
gburd, line 60: Looks like "300000000" changed to "420000000" and some formatting updated the whole chunk of text.  Did I miss anything?  Why that change?

db/dbformat.h
```diff
@@ -25,7 +25,8 @@
 static const int kNumOverlapLevels = 2;
 
 // Level-0 compaction is started when we hit this many files.
-static const size_t kL0_CompactionTrigger = 4;
+// Google:  static const size_t kL0_CompactionTrigger = 4;
+static const size_t kL0_CompactionTrigger = 6;
```
gburd, line 29: Why 6 not 4?

db/corruption_test.cc
```diff
@@ -323,7 +323,9 @@ class CorruptionTest {
   DBImpl* dbi = reinterpret_cast<DBImpl*>(db_);
 
   // Fill levels >= 1 so memtable compaction outputs to level 1
-  for (int level = 1; level < config::kNumLevels; level++) {
+  //  matthewv 1/10/14 - what does "levels" have to do with this,
```
gburd, line 326: Did you figure out "what levels had to do with that" or is this just a guess... it's odd wording that concerns me here.

gburd:
I've not built/run this code but by visual inspection it looks like improvements to heuristics in parts of LevelDB that make sense to me from a casual inspection.

db/version_set.cc
matthewvon, line 60: For once, I actually documented a change … see line 49 of comments above.  If still unclear, I will elaborate.

db/dbformat.h
matthewvon, line 29: wiki notes cover that … it is long story.

db/corruption_test.cc
matthewvon, line 326: I worked out the intent and realized someone made a booboo … see wiki notes.

matthewvon:
Thank you for the review.

db/version_set.cc
gburd, line 60: Here's the comment:
// level-0 file size of 420,000,000 applies to output files of this level
//   being written to level-1.  The value is 7 times the default maximum
//   write buffer size of 60,000,000.  Why seven times:  6 level-0 files typically compact
//   to one level-1 file and are each slightly larger than 60,000,000.

Which makes perfect sense, I thought it was another use of the magic "42" number rather than a real calculation.  Guess I was wrong!

gburd:
Without actually running the code (due to time constraints) I'm going to say I'm +1, but I'll suggest strongly that you should solicit at least one engineer to agree before merging it.

PR #120, opened 2014-02-13
Mv tuning6

matthewvon:
Detailed discussion here:  https://github.com/basho/leveldb/wiki/mv-tuning6

Creates limit on size of block cache based upon estimated optimal size for page cache.


andrewjstone:
Code looks good. Tests pass. :+1:

PR #122, opened 2014-02-23
Mmap size option

Licenser:
This pull requests add the mmap_size option to the options passable to the levedb instance. This is a more general approach to the developer_limit_memory flag which gives the developer control over how much space mmaped files will take. This is especially important in either development situates (as a substitution for developer_limit_memory) or when instead of one big leveldb file the application already spreads them around - it also allows to reserve bigger chunks then 20MB if it is helpful for the system.

### DIFF
--- a/db/db_impl.cc
+++ b/db/db_impl.cc
@@ -132,9 +132,11 @@ Options SanitizeOptions(const std::string& dbname,
   if (src.limited_developer_mem)
   {
       gMapSize=2*1024*1024L;
-      if (2*1024*1024L < result.write_buffer_size) // let unit tests be smaller
-          result.write_buffer_size=2*1024*1024L;
+  } else if (src.mmap_size) {
+      gMapSize=src.mmap_size;
   }   // if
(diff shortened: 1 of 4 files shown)


matthewvon:
We are in code freeze for Riak 2.0 release.  It will be a few weeks before I come back to review the code.  My apologies.

matthewvon:
New plan:  I am going to merge your option with mv-write-sizing branch.  I am leaving the default at 20M instead of the 1Mbyte in your branch.  This is because a recent test using 10Mbyte really dropped performance.  However, 20Mbyte default is no longer used for every file opened in mv-write-sizing.  

As of this moment the unit tests will not compile.  I will be fixing them tomorrow and merging your option then.

Please feel free to review the mv-write-sizing branch and leave comments here or in the branch.  There is one chunk of code in 3 places for determining recovery log sizing.  That will become a static function.

matthewvon:
PR #132 is now merged to "develop".  It contains the mmap_size variable.  eleveldb is still pending at this time.

PR #125, opened 2014-03-09
activate the AssertHeld() logic of Mutex and Spin classes. And address issue #100

matthewvon:
Scary to find that Mutex.AssertHeld() test, used throughout the leveldb code, performs no test.  This adds the necessary assert logic.  PR is for post-Riak 2.0 code freeze.

I used this code on a really long Bulk test sequence.  It did not find any mutex problems … thankfully.

This branch also contains an adjustment to VersionSet::LogAndApply().  Scott Fritchie created a test to randomly fail fail accesses.  That test found a flaw in the LogAndApply() logic flow that was adjusted for 2.0.  The logic flow now correctly differentiates between file failures and write failures … i.e. does not segfault.

### DIFF
--- a/db/memtable.cc
+++ b/db/memtable.cc
@@ -101,7 +101,7 @@ void MemTable::Add(SequenceNumber s, ValueType type,
   p += 8;
   p = EncodeVarint32(p, val_size);
   memcpy(p, value.data(), val_size);
-  assert((p + val_size) - buf == encoded_len);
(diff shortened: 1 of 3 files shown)


andrewjstone:
- [x] code inspection
- [x] make check

I couldn't get Scott's test in eleveldb to run, but it passed for him and Matthew. 

:+1:

PR #130, opened 2014-04-07
Mv tiered options

matthewvon:
Add options necessary to support tiered storage based upon the sst_? directories (introduced in Riak 1.3).

Branch discussion is here:
   https://github.com/basho/leveldb/wiki/mv-tiered-options

### DIFF
--- a/db/builder.cc
+++ b/db/builder.cc
@@ -36,7 +36,7 @@ Status BuildTable(const std::string& dbname,
 
   KeyRetirement retire(user_comparator, smallest_snapshot);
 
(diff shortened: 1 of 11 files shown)


gburd:
This set of changes is clean and straightforward, a natural extension of the work to separate level sst files into separate directories.  The changes have to do with partitioning sst levels into one of two (fast or slow) directories.

+1 After visual inspection and compile on CentOS with Erlang R17-rc2.  I will continue with a few longer running tests to quantify speed differences, but I don't think this pull should wait for that.  Please merge and test alongside other Riak 2.0 features.

PR #131, opened 2014-05-05
code to isolate Log() file flushes from happening within mutex...

matthewvon:
..._ locked regions.  Same with clean-up file deletes.

Detail discussion here: https://github.com/basho/leveldb/wiki/mv-mutex-cleanup

### DIFF
--- a/db/db_impl.cc
+++ b/db/db_impl.cc
@@ -287,9 +287,16 @@ void DBImpl::DeleteObsoleteFiles()
   //  simultaneous compaction
   if (RunningCompactionCount()<2)
   {
+      // each caller has mutex, we need to release it
+      //  since this disk activity can take a while
+      mutex_.AssertHeld();
+
       // Make a set of all of the live files
       std::set<uint64_t> live = pending_outputs_;
       versions_->AddLiveFiles(&live);
+
+      // release lock for disk work
+      mutex_.Unlock();
 
       // prune the database root directory
       std::vector<std::string> filenames;
@@ -310,6 +317,7 @@ void DBImpl::DeleteObsoleteFiles()
               KeepOrDelete(filenames[i], level, live);
           }   // for
       }   // for
+      mutex_.Lock();
   }   // if
 }
 
@@ -322,6 +330,8 @@ DBImpl::KeepOrDelete(
   uint64_t number;
   FileType type;
   bool keep = true;
+
+  // assumes mutex_ unlocked
 
   if (ParseFileName(Filename, &number, &type))
   {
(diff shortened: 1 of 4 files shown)


engelsanchez:
I'm starting to review this now.

util/posix_logger.h
```diff
@@ -83,7 +86,11 @@ class PosixLogger : public Logger {
       }
 
       assert(p <= limit);
-      fwrite(base, 1, p - base, file_);
+      // make fwrite single threaded (not fflush!)
```
engelsanchez: From reading about implementations of fwrite and fflush, it seems that they are both more likely than not already using locks to make the operations atomic. For weird platforms that do not use locks, I believe fflush would be unsafe to call from different threads.

matthewvon: I do not want to lock flush since it might take a long time.  I am ok removing the spin lock all together.  What would you prefer?

I had not noticed that the c standards now call for internal locking on fwrite() and fflush().

engelsanchez: I believe removing the lock altogether would be fine, as fwrite has its own lock on most platforms. If we ever want to run in a platform with thread unsafe I/O calls in the C runtime, we'll have no choice but to lock both. I think it is unlikely that we will (famous last words).

matthewvon: spin lock reverted … removed … etc.

engelsanchez: Just because I _love_ flogging a dead horse, here's a link to GNU's glibc fflush implementation with locking (I think condition on pthreads being linked): https://sourceware.org/git/?p=glibc.git;a=blob;f=libio/iofflush.c;h=5205d10604db395d9aa9f05b6b32991728644349;hb=HEAD#l40

db/db_impl.cc
```diff
@@ -323,6 +331,8 @@ DBImpl::KeepOrDelete(
   FileType type;
   bool keep = true;
 
+  // assumes mutex_ unlocked
+
```
engelsanchez, line 335: So the only thing I'm currently looking at is the use of versions_->LogNumber(), version_->PrevLogNumber() and versions_->ManifestFileNumber() below to determine whether to keep or delete log and manifes files.  Since now these values will be read outside of a lock, but are updated atomically in VersionSet::LogAndApply() which usually holds the DBImpl mutex passed in, I'm trying to convince myself whether values can be stale and whether that would lead to any problems.

engelsanchez, line 335: After talking to @matthewvon offline about the code using the Log and manifest numbers from version set without synchronization to determine when to keep or delete a log or manifest file if not needed anymore, we are convinced that stale reads here would at worst lead to a file being kept around a bit longer. Next time the routine is called, they would disappear.

engelsanchez:
I went over the code and discussed with Matthew. It seems all good to me. The one comment about unnecessary locking of the fwrite operation was addressed. :+1: :dancer: :boat:

PR #132, opened 2014-05-07
Mv write sizing

matthewvon:
All types of memory mapped files previously shared the same mapping size of 20Mbytes.  This is wasteful for many files, especially the smaller files created by Riak's AAE feature.  The waste is in two forms.  First the full 20Mbytes is always reserved by leveldb's flexcache.  This could lead to 2.5Gbytes of available memory being set aside unnecessarily for a node with 64 databases/vnodes/partitions.  The second waste is in physical writes to disk.  The entire 20Mbyte mapping is written to disk, then the file is resized smaller.  Again, this is a heavy burden for small files.

### DIFF
--- a/util/env_posix.cc
+++ b/util/env_posix.cc
@@ -282,11 +282,12 @@ class PosixMmapFile : public WritableFile {
  public:
   PosixMmapFile(const std::string& fname, int fd,
                 size_t page_size, size_t file_offset=0L,
-                bool is_async=false)
+                bool is_async=false,
+                size_t map_size=gMapSize)
       : filename_(fname),
         fd_(fd),
         page_size_(page_size),
-        map_size_(Roundup(gMapSize, page_size)),
+        map_size_(Roundup(map_size, page_size)),
         base_(NULL),
         limit_(NULL),
         dst_(NULL),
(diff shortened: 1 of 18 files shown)


util/env_posix.cc
```diff
@@ -775,6 +779,25 @@ class PosixEnv : public Env {
     {return(gImmThreads->m_WorkQueueAtomic + gWriteThreads->m_WorkQueueAtomic
           + gLevel0Threads->m_WorkQueueAtomic + gCompactionThreads->m_WorkQueueAtomic);};
 
+  virtual size_t RecoveryMmapSize(const struct Options * options) const
+    {
+      size_t map_size;
+
+      if (NULL!=options)
+      {
+        // large buffers, try for a little bit bigger than half hoping
```
andrewjstone, line 788: This code is the same as in db/builder.cc. Can it be abstracted to it's own function?

andrewjstone:
Besides my comment above, code looks good. Tests pass.

util/env_posix.cc
matthewvon, line 788: AARGH!  I missed this location.  I thought there were 3 places that needed to call the function.  You have found the 4th.  Bad Matthew.  Bad Bad Matthew.  Yes I will substitute the call.  Thank you for spotting this.

I did a copy/paste of the same code to multiple locations for proof of concept.  Missed this one on the clean-up.

andrewjstone:
:+1:

PR #139, opened 2014-07-03
Mv tuning9

matthewvon:
Make the decision to send file to page cache versus direct to disk dynamic.  Details here:

https://github.com/basho/leveldb/wiki/mv-tuning9

### DIFF
--- a/db/db_impl.cc
+++ b/db/db_impl.cc
@@ -1101,6 +1101,8 @@ Status DBImpl::OpenCompactionOutputFile(
   assert(compact != NULL);
   assert(compact->builder == NULL);
   uint64_t file_number;
+  bool pagecache_flag;
+
   {
     mutex_.Lock();
     file_number = versions_->NewFileNumber();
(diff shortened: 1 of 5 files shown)


engelsanchez:
I'm taking this review.

engelsanchez:
+1 on the code changes.

I'm curious about the fact that the logic to allow the new compaction output file to populate the page cache only cares about files in the lower levels fitting in the page cache.  That is, the OS might still evict lower level files before files from the same level sometimes. You would hope that its usage statistics would make it favor lower level files, but of course we don't have control over the algorithm in the OS. Do you think that issuing fadvise calls on already open files that this compaction file is replacing could have any positive effect?  That question is purely out of curiosity. 

That question is

matthewvon:
The current strategy for reading files is two part:  fadvise that the file is random read for normal operations (this prevents Linux from pre-reading wrong stuff into cache) and fadvise the file is sequential read for a compaction (now is a good time for Linux to pre-read).  If the file was read for compaction,  ~PosixRandomAccessFile() does exactly as you suggest and gives fadvise call to flush this file from page cache upon file close.

PR #140, opened 2014-07-09
Give Read and Iterator calls more consistent disk access on moderately loaded systems.

matthewvon:
Basho's leveldb tunings have emphasized write performance.  There are now times when the background compactions can consume the disk subsystem's capacity.  This leads to delayed Read and Iterator (2i) operations.  This branch spreads out background compactions on moderately loaded systems without sacrificing the total write throughput available for heavily load situations.

Discussion here:
https://github.com/basho/leveldb/wiki/mv-pacing


vinoski:
+1, looks good.

PR #142, opened 2014-08-21
Mv comparator segfault

matthewvon:


matthewvon:
previously verified by Scott

PR #143, opened 2014-11-26
Replace FILE* functions with their file descriptor equivalents

georgekola:
FILE\* functions are limited by FOPEN_MAX. On FreeBSD 10.1, this is
SHRT_MAX (32767). Unfortunately, on a busy server with lots of memory,
this limit could easily be hit. Note that open sockets and files opened
by open system call variants also count towards this limit. This
replacement is actually better for logging as we directly write to
file instead of fwrite followed by fflush.

### DIFF
--- a/util/env_posix.cc
+++ b/util/env_posix.cc
@@ -98,30 +98,45 @@ private:
 class PosixSequentialFile: public SequentialFile {
  private:
   std::string filename_;
-  FILE* file_;
+  int  fd_;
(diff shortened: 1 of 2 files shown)


matthewvon:
Basho has closed down.  Doubt anyone has access to merge the PR.  Just letting you know.

georgekola:
Thanks @matthewvon for letting me know

PR #146, opened 2014-12-24
Mv sequential tuning #146 [JIRA: RIAK-1477]

matthewvon:
The branch contains four independent changes, but all related to the customer issue:
- Create alternate throttle value for single vnode usage
- Restore the IsTrivialMove() function so that some compactions are simply renames
- Correct Options::mmap_size usage so that write_buffer is not forced to 20Mbytes all the time
- Implement concept of mv-compress-bypass branch in current code to eliminate compression attempts known to fail

Detail description: https://github.com/basho/leveldb/wiki/mv-sequential-tuning

### DIFF
--- a/README
+++ b/README
@@ -3,6 +3,8 @@ Authors: Sanjay Ghemawat (sanjay@google.com) and Jeff Dean (jeff@google.com)
 
 The original Google README is now README.GOOGLE.
 
+** Introduction
+
 This repository contains the Google source code as modified to benefit
 the Riak environment.  The typical Riak environment has two attributes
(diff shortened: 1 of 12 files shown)


matthewvon:
Found 2 warning messages when compiling with gcc 4.6 ... will fix and update.

Basho-JIRA:
Created this Jira story per Jon's request.  It appears as though MVM did the work and I don't believe it has been reviewed.  Please review and include in the 2.0.5 release. 

**_[posted via JIRA by Derek Somogyi]**_

andrewjstone:
:+1:

PR #150, opened 2015-03-10
Added explicit include of pthread.h

fadushin:
This fix allows leveldb to build cleanly on FreeBSD 10, which ships clang as the default compiler.


fadushin:
Covered by PR 151

PR #156, opened 2015-05-29
Set CC/C++ flags for FreeBSD

jj1bdx:
This fixes the C++ compiler name issue in FreeBSD 10.0 and later (no g++ installed).
This will _not_ break the compiler issues in FreeBSD 9.x and earlier since in those versions the stock c++ are also aliased to g++.
This will also fix the issue in https://github.com/basho/eleveldb/issues/97 but more thorough testing is required.


fadushin:
+1 add489f

jj1bdx:
@matthewvon Thx!

PR #161, opened 2015-06-17
Create method to dynamically enable / disable performance counters

matthewvon:
The leveldb performance counters now default to a disabled state. They are enabled when the file "/etc/riak/perf_counters" exists. The content of the file is irrelevant. leveldb's background thread polls for the file once a minute. This allows enabling and disabling the performance counters without restarting leveldb / Riak.

Detailed discussion is here:

   https://github.com/basho/leveldb/wiki/mv-perfcounter-disable

### DIFF
--- a/util/perf_count.cc
+++ b/util/perf_count.cc
@@ -527,7 +527,7 @@ PerformanceCounters * gPerfCounters(&LocalStartupCounters);
 
 
 
-    bool gPerfCountersDisabled=true;
+    volatile bool gPerfCountersDisabled=true;
     int PerformanceCounters::m_PerfSharedId=-1;
     int PerformanceCounters::m_LastError=0;
     volatile uint64_t PerformanceCounters::m_BogusCounter=0;
--- a/util/perf_count.cc
+++ b/util/perf_count.cc
@@ -52,7 +52,6 @@ namespace leveldb
 static PerformanceCounters LocalStartupCounters;
 PerformanceCounters * gPerfCounters(&LocalStartupCounters);
 
-
     SstCounters::SstCounters()
         : m_IsReadOnly(false),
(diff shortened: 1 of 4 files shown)


util/perf_count.cc
```diff
@@ -522,104 +525,106 @@ PerformanceCounters * gPerfCounters(&LocalStartupCounters);
...
+        {"GetVersion", true},
+        {"SearchLevel[0]", true},
+        {"SearchLevel[1]", true},
+        {"SearchLevel[2]", true},
+        {"SearchLevel[3]", true},
+        {"SearchLevel[4]", true},
+        {"SearchLevel[5]", true},
+        {"SearchLevel[6]", true},
+        {"TableCached", true},
+        {"TableOpened", true},
```
engelsanchez, line 561: Having used the TableOpened stat for that French customer to diagnose their really high latencies, I would say this one should remain by default. Don't you think?

matthewvon, line 561: I can argue the point both ways.
- yes, this is a counter that often says “out of memory” so performance is going to degrade
- no, we can get this counter immediately by having customer create /etc/riak/perf_counters file.

Open to you choosing.

Matthew

engelsanchez, line 561: Yes. Since we need a script that the customer needs to set up anyway and it's not exposed to stats yet, the point is moot. I'm fine with this staying like this and activating it along with all the others when needed.

engelsanchez:
:+1: 

I've reviewed the code and manually tested the mechanism. I did not re-run your performance benchmarks.

The location of the file is a bit hacky. The location of the check for the file is convenient but hacky. I'm sure I don't have to tell you this :). Hopefully we can make it more sophisticated when we work on exposing more of the stats to Riak.

PR #162, opened 2015-06-29
Increase calculated write throttle penalties on level-1 and decrease it on level-2.

matthewvon:
This branch tunes the changes made in mv-sequential-tuning branch to further enhance overall throughput of a heavy write load.  Throughput increased 11% to 29% in test loads.

Detail discussion of this change is found here:

https://github.com/basho/leveldb/wiki/mv-throttle-5


db/version_set.cc
```diff
@@ -1239,66 +1248,42 @@ VersionSet::UpdatePenalty(
...
-                penalty_score = (1.0<(static_cast<double>(level_bytes) / gLevelTraits[level].m_DesiredBytesForLevel)? 1.0 : 0.0);
-            }   // if
-
-            if (1.0<=penalty_score)
-                penalty+=(static_cast<int>(penalty_score));
-#else
-	    int loop, count, value, increment;
             const uint64_t level_bytes = TotalFileSize(v->files_[level]);
 
 	    count=static_cast<double>(level_bytes) / gLevelTraits[level].m_MaxBytesForLevel;
```
paulplace: Have tabs instead of spaces in the following lines: 1262, 1264, 1266, 1858, 1859, 1861

util/throttle.cc
```diff
@@ -140,6 +142,13 @@ ThrottleThread(
...
         }   // for
 
+        // capture current state of level-0 and other levels' backlog
+        gThrottleData[replace_idx].m_Backlog=gCompactionThreads->m_WorkQueueAtomic;
+        gPerfCounters->Add(ePerfThrottleBacklog1, gThrottleData[replace_idx].m_Backlog);
+
+        gThrottleData[0].m_Backlog=gLevel0Threads->m_WorkQueueAtomic;
+        gPerfCounters->Add(ePerfThrottleBacklog0, gThrottleData[0].m_Backlog);
+
 	// non-level0 data available?
```
paulplace: Have tabs instead of spaces in the following lines: 152, 176-180, 184, and 191

db/db_impl.cc
```diff
@@ -1549,8 +1554,8 @@ Status DBImpl::DoCompactionWork(CompactionState* compact) {
 
   if (status.ok()) {
     if (0!=compact->num_entries)
-        SetThrottleWriteRate((env_->NowMicros() - start_micros - imm_micros), compact->num_entries,
-                            is_level0_compaction, env_->GetBackgroundBacklog());
+        SetThrottleWriteRate((env_->NowMicros() - start_micros - imm_micros),
```
paulplace: imm_micros is not set to non-0 anywhere

paulplace:
Do we have any unit tests that drive the throttling calculations in isolation? Doing this effectively might require creating a mock version of the leveldb::Env class, perhaps something like InMemoryEnv class but writing to null instead of in-memory.

util/throttle.cc
```diff
@@ -30,6 +30,7 @@
 #include "util/cache2.h"
 #include "util/db_list.h"
 #include "util/flexcache.h"
+#include "util/hot_threads.h"
 #include "util/throttle.h"
 
 namespace leveldb {
```
paulplace, line 36: For lines 41-54, it would be helpful to add some comments about the manifest constants (e.g., why is THROTTLE_SCALING set to 17?), along with a description of how the gThrottledData array is used (e.g., that the first two slots in the array hold live data while the remaining slots hold rolling data from the past hour)

paulplace, line 36: For lines 38-39, I recommend changing from using raw pthread_mutex_t and pthread_cond_t raw types to the port::Mutex and port::CondVar types; then use scoped leveldb::MutexLock objects in place of explicit pthread_mutex_lock()/pthread_mutex_unlock() calls.

Then in lines 110-122, we would need to extend class CondVar, adding an overload of the Wait() method that takes a timeout parameter.

util/throttle.cc
```diff
@@ -140,6 +142,13 @@ ThrottleThread(
             tot_compact+=gThrottleData[loop].m_Compactions;
         }   // for
 
+        // capture current state of level-0 and other levels' backlog
+        gThrottleData[replace_idx].m_Backlog=gCompactionThreads->m_WorkQueueAtomic;
+        gPerfCounters->Add(ePerfThrottleBacklog1, gThrottleData[replace_idx].m_Backlog);
+
+        gThrottleData[0].m_Backlog=gLevel0Threads->m_WorkQueueAtomic;
```
paulplace, line 161: gThrottleData[0] is accessed in several places outside the lock, which means we could have an inconsistent state for gThrottleData[0] if someone calls SetThrottleWriteRate() for level 0 while we're taking care of business (e.g., on line 179); before releasing the lock on line 126, should make a copy of gThrottleData[0] and clear it (moving the memset on line 214) and then access the copy outside the lock

util/throttle.cc
```diff
@@ -168,6 +179,7 @@ ThrottleThread(
 	else if (0!=gThrottleData[0].m_Keys && 0!=gThrottleData[0].m_Compactions)
 	{
             pthread_mutex_lock(&gThrottleMutex);
+
             new_throttle=(gThrottleData[0].m_Micros / gThrottleData[0].m_Keys)
```
paulplace, line 193: We have a potential divide-by-0 here if someone calls SetThrottleWriteRate() after we check gThrottleData[0].m_Keys and gThrottleData[0].m_Compactions but before we lock gThrottleMutex.

(Never mind; I see that calls to SetThrottleWriteRate() only add to m_Keys and m_Compactions, so they won't be 0 until reset by the ThrottleThread() method).)

util/throttle.cc
paulplace, line 36: We have a minor race condition where if someone calls ThrottleShutdown() between the check of gThrottleRunning on line 94 and the locking of gThrottleMutex on line 107, the caller of ThrottleShutdown() will block for a full minute waiting to join the ThrottleThread; this can be eliminated by checking gThrottleRunning after locking gThrottleMutex on line 107.

db/version_set.cc
matthewvon: corrected.

util/throttle.cc
matthewvon: fixed

db/db_impl.cc
matthewvon: dead code removed

matthewvon:
No unit tests for throttling.  Please file as new “Issue”.

Matthew

util/throttle.cc
matthewvon, line 36: comments added

matthewvon, line 36: agreed.  would like more time to do that.  please file an Issue.

util/throttle.cc
matthewvon, line 161: gThrottleData[0].m_Backlog used to be a “live” variable.  This branch changes its definition.  It is now only set during this routine.  It is currently safe.

I would have a defensible position on gThrottleData[0].m_Keys and gThrottleData[0].m_Compactions being used unprotected in the else condition had they been declared as volatile.  But I have a vague memory of volatiles in a structure caused an issue on some compilers.  Maybe gcc wanted volatile at the struct level and some other compiler wanted it at the member variable level.  I just backed off.

This routine loops only once a minute and the mutex may get hit 60 to 120 times per second.  It is not a bottleneck.  Therefore moving discrete lock / unlock calls to encompass a larger portion of code.  These discrete calls can later be replaced by mutex objects.

util/throttle.cc
matthewvon, line 193: addressed by moving mutex lock to encompass larger portion of code.

util/throttle.cc
matthewvon, line 36: code changed as suggested

paulplace:
+1 code review looks good

matthewvon:
Thank you.

Next checkin has rebase of this branch, not new code. Need rebase for merge to succeed.

PR #163, opened 2015-06-29
Reactivate Google's PickLevelForMemTableOutput()

matthewvon:
This routine used to "be in the way" when efforts focused on random data ingest.  Now, it has new relevance as Riak's handoff and other sequential features come into play.

Detail discussion is here:

https://github.com/basho/leveldb/wiki/mv-initial-level-fix

### DIFF
--- a/db/db_impl.cc
+++ b/db/db_impl.cc
@@ -1863,7 +1863,9 @@ Status DBImpl::Write(const WriteOptions& options, WriteBatch* my_batch) {
 
 // REQUIRES: Writer list must be non-empty
 // REQUIRES: First writer must have a non-NULL batch
+// REQUIRES: mutex_ is held
 WriteBatch* DBImpl::BuildBatchGroup(Writer** last_writer) {
+  mutex_.AssertHeld();
   assert(!writers_.empty());
   Writer* first = writers_.front();
   WriteBatch* result = first->batch;
--- a/db/db_impl.cc
+++ b/db/db_impl.cc
@@ -735,13 +735,31 @@ Status DBImpl::WriteLevel0Table(volatile MemTable* mem, VersionEdit* edit,
             new_name=TableFileName(options_, meta.number, level);
             move_s=env_->RenameFile(old_name, new_name);
 
-            // argh!  logging while holding mutex ... cannot release
             if (move_s.ok())
+            {
+                // builder already added file to table_cache with 2 references and
+                //  marked as level 0 (used by cache warming) ... going to remove from cache
+                //  and add again correctly
+                table_cache_->Evict(meta.number, true);
+                meta.level=level;
+
+                // sadly, we must hold the mutex during this file open
+                //  since operating in non-overlapped level
+                Iterator* it=table_cache_->NewIterator(ReadOptions(),
+                                                       meta.number,
+                                                       meta.file_size,
+                                                       meta.level);
+                delete it;
+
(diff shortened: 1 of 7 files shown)


db/db_impl.cc
```diff
@@ -715,14 +715,51 @@ Status DBImpl::WriteLevel0Table(volatile MemTable* mem, VersionEdit* edit,
...
             std::string old_name, new_name;
 
             old_name=TableFileName(options_, meta.number, 0);
             new_name=TableFileName(options_, meta.number, level);
-            s=env_->RenameFile(old_name, new_name);
+            move_s=env_->RenameFile(old_name, new_name);
+
+            if (move_s.ok())
+            {
+                // builder already added file to table_cache with 2 references and
```
paulplace, line 740: The builder added the file to table_cache with mutex_ unlocked; is there any possibility that another thread could have referenced the file via the table_cache before we get to this point?

db/dbformat.h
```diff
@@ -44,7 +44,8 @@ static const size_t kL0_StopWritesTrigger = 12;
 // expensive manifest file operations.  We do not push all the way to
 // the largest level since that can generate a lot of wasted disk
 // space if the same key space is being repeatedly overwritten.
-static const int kMaxMemCompactLevel = 2;
+// Basho: push to kNumOverlapLevels +1 ... beyond "landing level"
+static const int kMaxMemCompactLevel = 3;
```
paulplace: The comment states that kMaxMemCompactLevel is kNumOverlapLevels +1, so why not just actually define it to be that value?

db/version_set.cc
```diff
@@ -475,9 +475,10 @@ bool Version::OverlapInLevel(int level,
 
 int Version::PickLevelForMemTableOutput(
     const Slice& smallest_user_key,
-    const Slice& largest_user_key) {
+    const Slice& largest_user_key,
+    const int level_limit) {
   int level = 0;
-#if 0
+#if 1
```
paulplace: Since this code is being put back into production, should remove the "#if 1/#endif" conditional compilation statements.

paulplace:
Notes to Matthew (github wouldn't let me add these comments inline, since they refer to parts of the code that were not modified):
- in db/db_impl.cc, line 1866: DBImpl::BuildBatchGroup() assumes mutex_ is locked; should probably add a comment and "mutex_.AssertHeld();" statement like we have with DBImpl::MakeRoomForWrite()
- db/version_set.cc, line 495: remember to fix signed/unsigned comparison warning (Matthew already caught this...just adding a reminder)

db/version_set.cc
matthewvon: corrected as recommended

matthewvon:
db/db_impl.cc line 1866:  updated as suggested

db/version_set.cc line 495 corrected:  now uses uint64_t instead of int64_t.

db/dbformat.h
matthewvon: changed as recommended.

db/db_impl.cc
matthewvon, line 740: The new file is not added to the manifest until after this routine returns.  The “edit” parameter gets set at line 767 with the new file number and its level.  No thread can access this file in the table cache because the file number is not known outside this routine.  The “edit” object subsequently gets posted to the manifest.  That is when the file number becomes visible.

paulplace:
+1 code review looks good

PR #164, opened 2015-06-29
Tool to rewrite a given sst file with new options, e.g. LZ4 compression

matthewvon:
This branch contains a skeleton tool for evaluating new .sst options.  The expectation is that the tool will be enhanced over time to provide more features and provide easy of use in migration / downgrade scenarios.  The branch includes the first feature for evaluation, LZ4 compression.

Detailed discussion here:

https://github.com/basho/leveldb/wiki/mv-sst-rewrite

### DIFF
--- a/table/format.cc
+++ b/table/format.cc
@@ -10,6 +10,7 @@
 #include "table/block.h"
 #include "util/coding.h"
 #include "util/crc32c.h"
+#include "util/lz4.h"
 #include "db/log_writer.h"
 
 namespace leveldb {
@@ -146,6 +147,28 @@ Status ReadBlock(RandomAccessFile* file,
                   if (!port::Snappy_Uncompress(data, n, ubuf)) {
                       s=Status::Corruption("corrupted compressed block contents");
                   }
+              }   // if
+
+              if (s.ok())
+              {
+                  delete[] buf;
+                  buf=NULL;
+                  result->data = Slice(ubuf, ulength);
+                  result->heap_allocated = true;
+                  result->cachable = true;
+              }   // if
+              break;
+          }
+
+          case kLZ4Compression: {
+              size_t ulength = DecodeFixed32(data);
+              size_t ret_val;
+              ubuf = new char[ulength];
+
+              ret_val=LZ4_decompress_safe(data+4, ubuf, n-4, ulength);
+              if (ret_val != ulength)
+              {
+                  s=Status::Corruption("corrupted LZ4 compressed block");
               }   // if
 
               if (s.ok())
--- a/table/table_builder.cc
+++ b/table/table_builder.cc
@@ -16,6 +16,7 @@
 #include "table/format.h"
 #include "util/coding.h"
 #include "util/crc32c.h"
+#include "util/lz4.h"
 
 namespace leveldb {
 
@@ -174,6 +175,8 @@ void TableBuilder::WriteBlock(BlockBuilder* block, BlockHandle* handle) {
   Slice block_contents;
   CompressionType type = r->options.compression;
   // TODO(postrelease): Support more compression options: zlib?
+  std::string * compressed;
+
   switch (type) {
     case kNoCompressionAutomated:
       // automation disabled compression
@@ -186,8 +189,8 @@ void TableBuilder::WriteBlock(BlockBuilder* block, BlockHandle* handle) {
       block_contents = raw;
       break;
 
-    case kSnappyCompression: {
-      std::string* compressed = &r->compressed_output;
+    case kSnappyCompression:
+      compressed = &r->compressed_output;
       if (port::Snappy_Compress(raw.data(), raw.size(), compressed) &&
           compressed->size() < raw.size() - (raw.size() / 8u)) {
         block_contents = *compressed;
@@ -199,7 +202,30 @@ void TableBuilder::WriteBlock(BlockBuilder* block, BlockHandle* handle) {
         r->sst_counters.Inc(eSstCountCompressAborted);
       }
       break;
-    }
+
+    case kLZ4Compression:
+      compressed = &r->compressed_output;
+      int limit, result_size;
+      limit=raw.size() - (raw.size() / 8u);
+
+      compressed->resize(limit+4);
+      result_size=LZ4_compress_default(raw.data(), (char *)(compressed->data())+4, raw.size(), limit);
+      if (result_size)
+      {
+          EncodeFixed32((char *)compressed->data(), raw.size());
+          compressed->resize(result_size+4);
+          block_contents = *compressed;
+      }
+      else {
+        // Snappy not supported, or compressed less than 12.5%, so just
+        // store uncompressed form
+        block_contents = raw;
+        type = kNoCompression;
+        r->sst_counters.Inc(eSstCountCompressAborted);
+      }
+      break;
+
+
   }
   WriteRawBlock(block_contents, type, handle);
   r->sst_counters.Add(eSstCountBlockWriteSize, block_contents.size());
--- a/tools/sst_rewrite.cc
+++ b/tools/sst_rewrite.cc
@@ -240,14 +240,16 @@ main(
                         error_seen=true;
                     }   // else
 
-                    for (it->SeekToFirst(); it->Valid() && s.ok(); it->Next())
+                    for (it->SeekToFirst();
+                         it->Valid() && s.ok() && builder->status().ok();
+                         it->Next())
                     {
                         leveldb::Slice key = it->key();
                         builder->Add(key, it->value());
                     }   // for
 
                     // hmmm, nothing new setting status right now.
-                    if (s.ok()) {
+                    if (s.ok() && builder->status().ok()) {
                         s = builder->Finish();
                     } else {
                         builder->Abandon();
--- a/tools/sst_rewrite.cc
+++ b/tools/sst_rewrite.cc
@@ -149,7 +149,7 @@ main(
     options.total_leveldb_mem=(512 << 20);
 
     // testing
-    options.block_size=16 << 10;
+//    options.block_size=16 << 10;
 
     for (cursor=argv+1; NULL!=*cursor && running; ++cursor)
     {
--- a/tools/sst_rewrite.cc
+++ b/tools/sst_rewrite.cc
@@ -320,16 +320,22 @@ main(
                         else
                         {
                             if (file1.GetSstCounter(leveldb::eSstCountKeys)==file2.GetSstCounter(leveldb::eSstCountKeys))
-                                fprintf(stderr, "%s, %s: Number of keys different.\n",
-                                        file1.GetFileName(), file2.GetFileName());
+                                fprintf(stderr, "%s, %s: Number of keys different, %ld vs %ld.\n",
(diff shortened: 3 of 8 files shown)


jadeallenx:
Is this the tool you're contemplating adding encryption support to eventually?

table/table_builder.cc
```diff
@@ -199,7 +202,30 @@ void TableBuilder::WriteBlock(BlockBuilder* block, BlockHandle* handle) {
         r->sst_counters.Inc(eSstCountCompressAborted);
       }
       break;
-    }
+
+    case kLZ4Compression:
```
jvoegele, line 206: It might be a larger refactoring than you're willing to make in the context of this PR, but it seems like this case statement and others that look very much like it could potentially be replaced with polymorphism to good effect.

matthewvon:
@mrallen1 Yes.  The initial purpose of this tool is to allow downgrading of LZ4 to Snappy and/or encrypted to non-encrypted .sst files ... i.e. give customers a way to access data in a prior version should something go wrong with initial version.

table/format.cc
```diff
@@ -159,6 +160,28 @@ Status ReadBlock(RandomAccessFile* file,
...
+
+              ret_val=LZ4_decompress_safe(data+4, ubuf, n-4, ulength);
+              if (ret_val != ulength)
+              {
+                  s=Status::Corruption("corrupted LZ4 compressed block");
+              }   // if
+
+              if (s.ok())
+              {
+                  delete[] buf;
```
paulplace, line 176: It would be a good idea to start using smart pointer wrapper classes, when possible, rather than ensuring that explicit deletes are included in all the right places. If std::unique_ptr is available on all platforms where Riak builds, that is the best approach to managing the lifetimes of resources allocated on the heap via new and new[]. If unique_ptr is not available, then you could use std::auto_ptr for single objects and write a similar helper class that deletes arrays of objects.

tools/sst_rewrite.cc
```diff
@@ -0,0 +1,353 @@
...
+
+    const leveldb::Status & GetStatus() const {return(m_LastStatus);};
+    const char * GetFileName() const {return(m_FileName.c_str());};
+
+    uint64_t GetSstCounter(unsigned Idx) const
+        {return(m_IsOpen ? m_TablePtr->GetSstCounters().Value(Idx) : 0);};
+
+protected:
+    leveldb::Options & m_Options;
+    std::string m_FileName;
```
paulplace: This can probably be declared as const. I try to make as many data members as possible const, then I remove constness (or make them mutable) as necessary.

tools/sst_rewrite.cc
```diff
@@ -0,0 +1,353 @@
...
+LDbTable::LDbTable(
+    leveldb::Options & Options,
+    std::string & FileName)
+    : m_Options(Options), m_FileName(FileName),
+      m_FilePtr(NULL), m_TablePtr(NULL), m_FileSize(0), m_IsOpen(false)
+{
+    m_LastStatus=m_Options.env->GetFileSize(m_FileName, &m_FileSize);
+
+    if (m_LastStatus.ok())
+        m_LastStatus=m_Options.env->NewRandomAccessFile(m_FileName, &m_FilePtr);
```
paulplace: I prefer curly braces around single conditional statements: "if (bool) {statement;}" instead of "if (bool) statement;"

This avoids the potential bug of someone adding a second statement later and forgetting to wrap the two in braces.

tools/sst_rewrite.cc
```diff
@@ -0,0 +1,353 @@
...
+        // use fadvise to start file pre-read
+        m_FilePtr->SetForCompaction(m_FileSize);
+    }   // if
+
+    m_IsOpen=m_LastStatus.ok();
+
+    if (!m_IsOpen)
+    {
+        // some people would throw() at this point, but not me
+        delete m_TablePtr;
```
paulplace: This clean up code is the same as in the dtor; both could be moved to a common private routine.

tools/sst_rewrite.cc
```diff
@@ -0,0 +1,353 @@
...
+    bool error_seen, running, compare_files;
+    char ** cursor;
+
+    compare_files=false;
+    error_seen=false;
+    running=true;
+
+    // Options: needs filter & total_leveldb_mem initialized
+    leveldb::Options options;
+    options.filter_policy=leveldb::NewBloomFilterPolicy2(16);
```
paulplace, line 162: I'm not sure what the parameter value 16 signifies (same goes for the (512 << 20) on the next line). I recommend either replacing these with symbolic constants of some sort (e.g., enumeration, const integer, or even a C-style macro) or adding a comment indicating the meaning of the number.

tools/sst_rewrite.cc
```diff
@@ -0,0 +1,353 @@
...
+    options.total_leveldb_mem=(512 << 20);
+
+    // testing
+//    options.block_size=16 << 10;
+
+    for (cursor=argv+1; NULL!=*cursor && running; ++cursor)
```
paulplace: There is no check of argc here; is argv+1 guaranteed to be meaningful if argc is 1?

paulplace: This for loop checks the running boolean; it should either check the error_seen boolean as well, or places that set error_seen should set running=false; if the intent is to continue processing command line args even though an error was seen, a comment to this effect should be added just above this for loop

tools/sst_rewrite.cc
```diff
@@ -0,0 +1,353 @@
...
+        }   // if
+
+        // sst file
+        else
+        {
+            std::string fname;
+            fname=*cursor;
+
+            // do a rewrite
+            if (!compare_files)
```
paulplace, line 217: The notion of chaining several actions together by allowing this loop to process a new option on each pass is interesting, but it leads to some possible confusion. For example, "sst_rewrite -c file1 file2" gives a different result (compare file1 to file2) than "sst_rewrite file1 file2 -c" (write new options to file1 and file2). Perhaps requiring an explicit option on each pass would help. For example, requiring that -c or -w be explicitly listed before the file name(s) will make clear which action will be performed.

tools/sst_rewrite.cc
```diff
@@ -0,0 +1,353 @@
...
+    {
+        // option flag?
+        if ('-'==**cursor)
+        {
+            char flag;
+
+            flag=*((*cursor)+1);
+            switch(flag)
+            {
+                //case 'b':  options.block_size=4096; break;
```
paulplace: Should probably add in some initial options to make this utility useful. Setting the block size and setting the compression method would both be good additions.

tools/sst_rewrite.cc
```diff
@@ -0,0 +1,353 @@
...
+                        builder = new leveldb::TableBuilder(options, outfile);
+                    else
+                    {
+                        // Table::Open failed on file "fname"
+                        fprintf(stderr, "%s: NewWritableFile failed (%s)\n",
+                                fname.c_str(), s.ToString().c_str());
+                        error_seen=true;
+                    }   // else
+
+                    for (it->SeekToFirst(); it->Valid() && s.ok(); it->Next())
```
paulplace: I recommend adding a call to builder->status.ok() in the conditional section of the for loop. Similarly, I suggest adding the same check on line 216.

tools/sst_rewrite.cc
```diff
@@ -0,0 +1,353 @@
...
+        // sst file
+        else
+        {
+            std::string fname;
+            fname=*cursor;
+
+            // do a rewrite
+            if (!compare_files)
+            {
+                leveldb::WritableFile *outfile = NULL;
```
paulplace: All of these objects that are allocated on the heap can be wrapped in std::auto_ptr<> objects and the explicit deletes below can be removed. If std::unique_ptr<> is available on all platforms, then that is preferred to auto_ptr, which is now deprecated.

tools/sst_rewrite.cc
```diff
@@ -0,0 +1,353 @@
...
+                                fprintf(stderr, "%s, %s: Walk of keys terminated early (%d, %d).\n",
+                                        file1.GetFileName(), file2.GetFileName(),
+                                        (int)it1->Valid(), (int)it2->Valid());
+                                error_seen=true;
+                            }
+                        }   // if
+                        else
+                        {
+                            if (file1.GetSstCounter(leveldb::eSstCountKeys)==file2.GetSstCounter(leveldb::eSstCountKeys))
+                                fprintf(stderr, "%s, %s: Number of keys different.\n",
```
paulplace: Several of these error messages can include more information, which would be useful. For example, this message could include the number of keys in each file.

tools/sst_rewrite.cc
matthewvon: agreed.  done.  available next check-in.

tools/sst_rewrite.cc
matthewvon: trying it.  unfortunately it confuses the auto alignment / tabbing of that line in my emacs.  hmm. done.  available next check-in.

tools/sst_rewrite.cc
matthewvon: done.  but placed in protected out of old MFC habit

tools/sst_rewrite.cc
matthewvon, line 162: comments added

tools/sst_rewrite.cc
matthewvon: error_seen added

matthewvon: argv is an array with trailing NULL entry.  so if argc=1, then NULL==argv+1 prior to loop.

tools/sst_rewrite.cc
matthewvon, line 217: compare is developer feature.  put in a minor safety check.  users would be activating -w, the default, and likely via Basho provided downgrade script.  not adding armor plating at this time.  Though your case is valid.

tools/sst_rewrite.cc
matthewvon: added as suggested.

tools/sst_rewrite.cc
matthewvon: actually, while this is a good suggestion ... it will not work with existing Google code ... not in the mood to change:

./include/leveldb/table_builder.h:81:8: error: ‘bool leveldb::TableBuilder::ok() const’ is private

tools/sst_rewrite.cc
matthewvon: std::unique_ptr not even available on my stock debian.  Likely a package would fix that, but means all Basho developers would have to add the package ... not in the mood for that fight.  Used auto_ptr on all but outfile.  A WriteableFile *\* must be passed into NewWritableFile and my programming foo was not strong enough to make that happen.

tools/sst_rewrite.cc
matthewvon: Done.

tools/sst_rewrite.cc
paulplace: leveldb::TableBuilder::ok() is private, but leveldb::TableBuilder::status() is not. My suggested code was not quite right; it should be "builder->status().ok()".

paulplace:
+1 code review looks good

PR #165, opened 2015-07-30
Mv cache object warming  [JIRA: RIAK-2350]

matthewvon:
This branch adds code to create a temporary file contain the list of .sst table files at the time of shutdown.  The next start up will open all of the .sst table files as part of the database Open operation.  This preload of the file cache greatly reduces user Get and Iterate request latencies during the early minutes of a leveldb restart.

Detail here:  https://github.com/basho/leveldb/wiki/mv-cache-object-warming


db/table_cache.cc
```diff
@@ -197,4 +201,144 @@ TableCache::GetStatisticValue(
...
+ *  and preload them into the table cache
+ */
+void
+TableCache::PreloadTableCache()
+{
+    struct LogReporter : public log::Reader::Reporter {
+        Env* env;
+        Logger* info_log;
+        const char* fname;
+        Status* status;  // NULL if options_.paranoid_checks==false
```
paulplace: the "NULL if options_.paranoid_checks==false" comment does not apply here

db/table_cache.cc
```diff
@@ -197,4 +201,144 @@ TableCache::GetStatisticValue(
...
+    Status s;
+    std::string cow_name;
+    WritableFile * cow_file;
+    log::Writer * cow_log;
+
+    cow_name=CowFileName(dbname_);
+    s = env_->NewWritableFile(cow_name, &cow_file, 4*1024L);
+    if (s.ok())
+    {
+        std::string buffer;
```
paulplace: I recommend setting the initial capacity of the buffer variable (say, to 4KB) to avoid a bunch of re-allocs as cache file records are added to the std::string.

db/table_cache.cc
```diff
@@ -197,4 +201,144 @@ TableCache::GetStatisticValue(
...
+
+        while (cow_log->ReadRecord(&record, &buffer) && s.ok())
+        {
+            Slice input = record;
+            uint32_t tag, level;
+            uint64_t file_no, file_size;
+            Cache::Handle * handle;
+            Status s2;
+
+            while (GetVarint32(&input, &tag))
```
paulplace: I recommend this set of GetVarintXx() calls be moved to cache2.cc as a companion to the SharedLRUCache2::WriteFileCacheObjectWarming() method, which writes the serialized record. The serialization/deserialization logic should be kept together, if at all possible.

You could declare a structure whose data members are level, file_no, and file_size, and pass a reference to a std::list<> of this structure that the new method would populate from the input Slice object.

Then this method would change from iterating over the GetVarintXx() calls to iterating over the std::list<>, calling FindTable() for each entry in the list.

db/table_cache.cc
matthewvon: Corrected.

db/table_cache.cc
matthewvon: Done.

db/table_cache.cc
matthewvon: How about we keep it simple and just add comments in the decode section that says where encode happens, and comments in encode section where decode happens.  Bouncing the code to a static function in cache2 does put read and write together, but seems like overkill.  Do you feel strongly about this?

db/table_cache.cc
```diff
@@ -197,4 +201,150 @@ TableCache::GetStatisticValue(
...
+        {
+            Slice input = record;
+            uint32_t tag, level;
+            uint64_t file_no, file_size;
+            Cache::Handle * handle;
+            Status s2;
+
+            // the on disk format is created in WriteFileCacheObjectWarming()
+            //  (util/cache2.cc)
+            while (GetVarint32(&input, &tag))
```
matthewvon: Ok, having second thoughts about this "where do we put the format code read and write".  Just realized that there is another "write" location in tools/cache_warm.cc.  Going to rethink this.  Likely going to make an object.  Then put read and write to std::string within that object.  Should make everyone happy.

bashopatricia:
create jira issue

util/cache2.cc
```diff
@@ -421,6 +423,33 @@ class ShardedLRUCache2 : public Cache {
 
   } // ShardedLRUCache2::PurgeExpiredFiles
 
+  // Walk all cache entries, calling functor Acc
+  //  Writes list of open files to log record
```
paulplace: The "Writes list of open files to log record" looks to be a holdover from the previous COW-specific implementation

paulplace:
+1 f3a81b4

paulplace:
+1 a26cff8

Basho-JIRA:
1/21 per MVM, not ready for this release.  2.2 or 2.0.7 

PR #171, opened 2015-09-11
pp hot threads backport

paulplace:
This branch is the leveldb work needed to support backporting the hot_threads class to eleveldb.

The corresponding pull request for eleveldb is https://github.com/basho/eleveldb/pull/147.

The thread pool in eleveldb, eleveldb_thread_pool, contains a race condition in the job hand-off code that is the subject of at least one bug report (https://github.com/basho/eleveldb/issues/141). This race condition was addressed in the hot_threads code added to leveldb (https://github.com/basho/leveldb/wiki/mv-hot-threads).

Another bit of refactoring in this branch is to split the reference-counted object base class into a separate class, RefObjectBase, so that it can be used in both eleveldb/c_src/refobjects.h and leveldb/util/thread_tasks.h.

Other work done in this branch supports replacing all pthread_mutex_t and pthread_cond_t variables with the corresponding wrapper classes in the port namespace. One other change included in this branch is to re-enable naming the current thread, which makes core dumps a bit more readable.

Note that a stretch goal for this branch was to refactor the queue monitor thread in hot_threads, removing the need for a semaphore; this work did not happen.


paulplace:
A couple more minor comments about the code in hot_threads.cc:
- pthread_create() loop in HotThreadPool ctor should break in the "else" case to avoid possible loss of ret_val
- check of shutdown_pending() in HotThreadPool::FindWaitingThread() does not short-circuit the do/while loop

util/throttle.cc
```diff
@@ -283,10 +298,20 @@ void ThrottleShutdown()
...
+
+        // lock gThrottleMutex so that we can signal gThrottleCond
+        {
+            MutexLock lock(gThrottleMutex);
+            gThrottleCond->Signal();
+        } // unlock gThrottleMutex
+
         pthread_join(gThrottleThreadId, NULL);
+
+        delete gThrottleCond;
```
matthewvon: Problem:  see util/env_posix.cc function Env::Shutdown().  The throttle object is deleted before the compaction threads.  So routine SetThrottleWriteRate() could still be called after the Condition variable and mutex get deleted / set to NULL.  I feel a segfault.

The question becomes whether or not the Throttle could be deleted after the threads shutdown.  Thinking on that right now.

port/port_posix.cc
```diff
@@ -48,6 +61,19 @@ void CondVar::Wait() {
   PthreadCall("wait", pthread_cond_wait(&cv_, &mu_->mu_));
 }
 
+bool CondVar::Wait(struct timespec* pTimespec) {
+  int result = pthread_cond_timedwait(&cv_, &mu_->mu_, pTimespec);
+  if (0 == result) {
+    return true; // condition variable signaled
```
matthewvon: Did I ever explain why I really, really hate multiple "return" statements in a function?

matthewvon:
sst_rewrite removed.  gThrottleRunning now volatile.

PR #172, opened 2015-09-21
Mv cache warming tool  [JIRA: RIAK-2351]

matthewvon:
Internal / consulting tool for creating an initial cache warming file.  This is NOT industrial grade and requires hand adjustments to shell script depending upon the user environment.

### DIFF
--- a/tools/cache_warm.cc
+++ b/tools/cache_warm.cc
@@ -1,0 +1,278 @@
+// -------------------------------------------------------------------
+//
+// cache_warm.cc
+//
+// Copyright (c) 2015 Basho Technologies, Inc. All Rights Reserved.
+//
+// This file is provided to you under the Apache License,
+// Version 2.0 (the "License"); you may not use this file
+// except in compliance with the License.  You may obtain
+// a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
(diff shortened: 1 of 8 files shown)


matthewvon:
Original comments from Paul Place's code review:

I didn't apply the same standards as I would to production code, since my understanding is that the purpose of this tool is to ensure the new code in mv-cache-object-warming has a COW file the first time it runs after upgrading to this branch.

All looks fine except for the following in tools/cache_warm.cc
    line 129: NAME_MAX does not include the terminating null character, so should make it file[NAME_MAX+1]
    line 264: need to update the comman_help() method from sst_rewrite to cache_warm

--Paul

tools/cache_warm.cc
```diff
@@ -0,0 +1,280 @@
...
+        }   // if
+
+        // file of .sst pathnames
+        else
+        {
+            FILE * list;
+            bool ignore_file;
+            std::string cur_path, record;
+            size_t line_count;
+            char buffer[PATH_MAX], vnode[PATH_MAX], file[NAME_MAX], * chr_ptr;
```
matthewvon: Paul Place writes: NAME_MAX does not include the terminating null character, so should make it file[NAME_MAX+1]

tools/cache_warm.cc
```diff
@@ -0,0 +1,280 @@
...
+    if (1==argc)
+        command_help();
+
+    return( error_seen ? 1 : 0 );
+
+}   // main
+
+
+void
+command_help()
```
matthewvon, line 263: Paul Place writes:  need to update the comman_help() method from sst_rewrite to cache_warm

bashopatricia:
create jira issue

Basho-JIRA:
1/21 per MVM, not ready for this release. 2.2 or 2.0.7 

PR #173, opened 2015-09-30
Mv timed grooming

matthewvon:
This branch initiates smaller compaction sets in levels 0 and 1 based upon elapsed time.  This is to improve read latencies in datasets with small to medium write loads.  

See https://github.com/basho/leveldb/wiki/mv-timed-grooming for details.

### DIFF
--- a/db/dbformat.h
+++ b/db/dbformat.h
@@ -31,6 +31,8 @@ static const size_t kL0_CompactionTrigger = 6;
 // Level-0 (any overlapped level) number of files where a grooming
 //     compaction could start
 static const size_t kL0_GroomingTrigger = 4;
+static const size_t kL0_GroomingTrigger10min = 2;
+static const size_t kL0_GroomingTrigger20min = 1;
 
 // Soft limit on number of level-0 files.  We slow down writes at this point.
 static const size_t kL0_SlowdownWritesTrigger = 8;
--- a/db/dbformat.h
+++ b/db/dbformat.h
@@ -34,6 +34,10 @@ static const size_t kL0_GroomingTrigger = 4;
 static const size_t kL0_GroomingTrigger10min = 2;
 static const size_t kL0_GroomingTrigger20min = 1;
 
+// ... time limits in microseconds
+static const size_t kL0_Grooming10minMicros = 10 * 60 * 1000000;
+static const size_t kL0_Grooming20minMicros = 20 * 60 * 1000000;
+
 // Soft limit on number of level-0 files.  We slow down writes at this point.
 static const size_t kL0_SlowdownWritesTrigger = 8;
 
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -1086,9 +1086,12 @@ VersionSet::Finalize(Version* v)
     double best_score = -1;
     bool compaction_found;
     bool is_grooming;
+    uint64_t micros_now;
 
     compaction_found=false;
     is_grooming=false;
+    micros_now=env_->NowMicros();
+
     for (int level = v->compaction_level_+1; level < config::kNumLevels-1 && !compaction_found; ++level)
     {
         bool compact_ok;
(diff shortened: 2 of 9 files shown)


db/dbformat.h
```diff
@@ -31,6 +31,8 @@ static const size_t kL0_CompactionTrigger = 6;
 // Level-0 (any overlapped level) number of files where a grooming
 //     compaction could start
 static const size_t kL0_GroomingTrigger = 4;
+static const size_t kL0_GroomingTrigger10min = 2;
```
paulplace, line 34: Can remove "static" from all these constants

db/version_set.cc
```diff
@@ -1125,6 +1128,22 @@ VersionSet::Finalize(Version* v)
         // consider this level
         if (compact_ok)
         {
+            size_t grooming_trigger;
+
+            // which grooming trigger point?  based upon how long
+            //  since last compaction on this level
+            //   - less than 10 minutes?
+            if (micros_now<(m_CompactionStatus[level].m_LastCompaction + (1000000 * 60 *10)))
```
paulplace: I recommend pre-computing the elapsed time as "uint64_t elapsed_micros = micros_now - m_CompactionStatus[level].m_LastCompaction;" so that the comparisons become:

```
uint64_t elapsed_micros = micros_now - m_CompactionStatus[level].m_LastCompaction;
if (elapsed_micros < (1000000 * 60 *10))
    ...
else if (elapsed_micros < (1000000 * 60 *20))
```

db/version_set.cc
```diff
@@ -1125,6 +1128,22 @@ VersionSet::Finalize(Version* v)
...
+                grooming_trigger=config::kL0_GroomingTrigger;
+
+            //   - less than 20 minutes?
+            else if (micros_now<(m_CompactionStatus[level].m_LastCompaction + (1000000 * 60 *20)))
```
paulplace: I recommend defining (1000000 \* 60 *10) and (1000000 \* 60 *20) as constants (probably in db/dbformat.h, along with kL0_GroomingTrigger10min and kL0_GroomingTrigger20min)

db/version_set.h
```diff
@@ -374,9 +375,10 @@ class VersionSet {
   {
       bool m_Submitted;     //!< level submitted to hot thread pool
       bool m_Running;       //!< thread actually running compaction
+      uint64_t m_LastCompaction; //!<NowMicros() when last compaction completed
 
       CompactionStatus_s()
-      : m_Submitted(false), m_Running(false)
+      : m_Submitted(false), m_Running(false), m_LastCompaction(Env::Default()->NowMicros())
```
paulplace: The problem with initializing m_LastCompaction via Env::Default()->NowMicros() is that you're mixing and matching environment objects, since m_LastCompaction is set to env_->NowMicros() in SetCompactionDone() above. Normally this assumption is fine, but a different environment object might be used in unit tests. Should this be initialized to 0 here?

db/dbformat.h
matthewvon, line 34: I do not disagree.  But "static const" is part of the original Google writing.  I see no benefit to removing.  Is there a benefit beyond removing redundancy of declaration?

db/version_set.cc
matthewvon: I like the thought.  The thing that bothers me is that clock is not guaranteed to move forward on platforms forced to use gettimeofday().  So micros_now could be smaller than m_LastCompaction.  Going to add test for the case and force elapsed to zero.

db/version_set.h
matthewvon: Ok, initializing to zero.  Change test in version_set.cc to explicitly check for value of zero and do right thing.

db/dbformat.h
paulplace, line 34: It depends on the compiler and linker, but I have seen static variables that are declared and instantiated in a header file cause an instance of the variable to be statically instantiated in each source file that includes the header file. It's not a big deal since the variables are also const, but it does add extra space by duplicating the variables in multiple source files.

In this case, I am cool with not changing anything.

matthewvon, line 34: Yeah, let this one slide.

db/version_set.h
paulplace: Hmm...I think the "initialize to 0" is in the wrong place. Should be in the ctor for CompactionStatus_s. With this change, SetCompactionDone() will always set m_LastCompaction to 0, which means the calculation in VerstionSet::Finalize() will always set elapsed_micros to 0.

matthewvon: oops

paulplace:
+1 code review looks good

PR #175, opened 2015-10-22
Cleaned up most but not all clang warnings.

fadushin:
This PR removes some unused code that Clang identifies in its build of the leveldb source tree.

It also changes the forward declaration in env.h of the Options type from class to struct, which remove the majority of the warnings from the build.

Still using deprecated semaphore APIs on Darwin, but that is a more extensive change to make.


matthewvon:
+1

PR #176, opened 2015-11-14
pp bugfix freebsd compiler issues

paulplace:
The pp-bugfix-freebsd-compiler-issues branch fixes an issue found only when compiling on FreeBSD. The FreeBSD build machine build-freebsd-9.2.bos1 uses a fairly old version of gcc, which chokes when compiling recent changes to db/skiplist.h.

The issue with building on FreeBSD was reported here: https://github.com/basho/riak_pipe/pull/102#issuecomment-156502047

While in the leveldb code on FreeBSD, I also fixed a compiler warning about 'signed vs. unsigned' comparisons in db/db_impl.cc.


matthewvon:
+1 Merging

PR #182, opened 2015-12-31
Mv ldb config h

matthewvon:
Create include/leveldb/ldb_config.h to hold compile time #defines.  This allows external code to use some of the leveldb internal functions that Basho has augmented.

Full branch discussion is available here:

https://github.com/basho/leveldb/wiki/mv-ldb-config-h


paulplace:
+1 daef7d2

PR #183, opened 2016-01-04
Cleanup of Makefile for ongoing developer use

matthewvon:
After 3 years, finally decided to clean up the Makefile for daily developer use.  Was originally written for single-shot library builds.

Details of changes are here:  https://github.com/basho/leveldb/wiki/mv-make-makeover


Makefile
```diff
@@ -8,6 +8,7 @@ INSTALL_PATH ?= $(CURDIR)
 #-----------------------------------------------
```
paulplace, line 8: Since you've made substantive changes, it might also be a good time to update the copyright notice. Or maybe not. I'm not sure about Basho's policies on such things.

matthewvon:
On Jan 5, 2016, at 5:11 PM, Paul Place pplace@basho.com wrote:

The leveldb make changes look good. I asked about updating the copyright date for the Makefile, but all else is good: builds on OS X, tests run, etc.

--Paul

PR #184, opened 2016-01-08
Startup segfault/race condition [JIRA: RIAK-2349]

matthewvon:
Move the AddDB() call to point after all object initialization complete to eliminate race condition that leads to segfault.

Detail description here:

https://github.com/basho/leveldb/wiki/mv-startup-segfault

### DIFF
--- a/db/db_impl.cc
+++ b/db/db_impl.cc
@@ -192,7 +192,6 @@ DBImpl::DBImpl(const Options& options, const std::string& dbname)
       block_size_changed_(0), last_low_mem_(0)
 {
   current_block_size_=options_.block_size;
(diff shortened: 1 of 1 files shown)


bashopatricia:
create jira issue

Basho-JIRA:
Per Matthew - Eleveldb tag 2.0.11 addresses PR 184.  It can be used in all Riak releases.  

matthewvon:
The customer reporting the segfault has confirmed the fix.

Basho-JIRA:
Once we update the dependencies this should be closed as well. 

**_[posted via JIRA by Douglas Rohrer]**_

PR #185, opened 2016-02-01
Mv expiry fallback

matthewvon:
Create a release without expiry support that can process expiry enabled files.  This is to be a fallback release should any portion of Riak fail and require a fallback to a prior release.

### DIFF
--- a/db/dbformat.h
+++ b/db/dbformat.h
@@ -62,7 +62,9 @@ class InternalKey;
 // data structures.
 enum ValueType {
   kTypeDeletion = 0x0,
-  kTypeValue = 0x1
+  kTypeValue = 0x1,
+  kTypeValueWriteTime = 0x2,
+  kTypeValueExplicitExpiry = 0x3
 };
 // kValueTypeForSeek defines the ValueType that should be passed when
 // constructing a ParsedInternalKey object for seeking to a particular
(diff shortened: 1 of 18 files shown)


db/dbformat.h
```diff
@@ -200,8 +245,12 @@ inline bool ParseInternalKey(const Slice& internal_key,
   unsigned char c = num & 0xff;
   result->sequence = num >> 8;
   result->type = static_cast<ValueType>(c);
-  result->user_key = Slice(internal_key.data(), n - 8);
-  return (c <= static_cast<unsigned char>(kTypeValue));
+  if (IsExpiryKey((ValueType)c))
+    result->expiry=DecodeFixed64(internal_key.data() + n - 16);
```
paulplace: Would be good to get rid of hard-coded "- 16", much like most of the "- 8"s were replaced.

matthewvon: Done.

db/dbformat.h
```diff
@@ -107,18 +111,59 @@ extern void AppendInternalKey(std::string* result,
 extern bool ParseInternalKey(const Slice& internal_key,
                              ParsedInternalKey* result);
 
+inline ValueType ExtractValueType(const Slice& internal_key) {
+  assert(internal_key.size() >= 8);
+  const size_t n = internal_key.size();
+  unsigned char c = DecodeLeastFixed64(internal_key.data() + n - 8);
```
paulplace: Can this hardcoded "- 8" be replaced with something more descriptive?

matthewvon: done

paulplace:
+1 c56a2b6

PR #186, opened 2016-02-24
Fixed infinite loop seen in a unit test on FreeBSD

paulplace:
I reworked the looping logic in Compaction::CalcInputStats() (in db/version_set.cc) to avoid an infinite loop we saw in unit tests on FreeBSD.

The version of STL on FreeBSD was storing the Compaction::inputs_ array of two std::vector objects in such a way that the logic in Compaction::CalcInputStats() that was attempting to walk both levels of input files would never terminate. This happened because inputs_[0].end() matched inputs_[1].begin().

Now we perform a traditional loop that does not use std::vector::iterator.

Note that the diffs look a bit ugly because the re-write removed a nesting level of a large block of code. The un-indented code is otherwise unchanged. You can see a cleaner view of the changes by ignoring whitespace-only differences; github allows you to do this by appending "?w=1" to the URL: https://github.com/basho/leveldb/pull/186/files?w=1

### DIFF
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -1928,7 +1928,6 @@ Compaction::CalcInputStats(
 {
     uint64_t temp, temp_cnt;
     size_t value_count, key_count, block_count;
-    std::vector<FileMetaData *>::iterator it;
 
     if (!stats_done_)
     {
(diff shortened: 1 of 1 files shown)


matthewvon:
+1 to merge

PR #187, opened 2016-02-25
Mv expiry [JIRA: RIAK-2573]

matthewvon:
not ready for prime time ... yet

Detailed description here:  https://github.com/basho/leveldb/wiki/mv-expiry


include/leveldb/iterator.h
```diff
@@ -64,6 +65,10 @@ class Iterator {
   // REQUIRES: !AtEnd() && !AtStart()
   virtual Slice value() const = 0;
 
+  // Riak specific:  if a database iterator, returns key meta data
+  // REQUIRES: Valid()
+  virtual KeyMetaData & keymetadata() const {return(keymetadata_); };
```
paulplace, line 70: keymetadata() is a const method, so it should probably return a const reference (unless there's some expected usage that requires non-const).

matthewvon, line 70: Note that keymetadata_ is mutable.  So this is where things got ugly.  Iterators are const everywhere useful.  Expiry creates the need to return not only the value, but also the KeyMetaData.  I considered several alternatives.  The one that created fewer code changes was to keep the const Iterator but give it a mutable storage for KeyMetaData.  Open to alternate construction ideas.

paulplace, line 70: I'm not suggesting you lose mutability of the `keymetadata_` member; I get why you did that. I was suggesting that the returned reference could also be const unless otherwise needed. I changed the declaration to

```
virtual const KeyMetaData & keymetadata() const {return(keymetadata_); };
```

in my local copy and everything still builds and runs just fine. I'm just not sure if you expect the need to return a non-const reference to `keymetadata_` in the future.

db/c_test.c
```diff
@@ -11,8 +13,13 @@
 #include <sys/types.h>
 #include <unistd.h>
 
+using leveldb::ValueType;
+
+struct leveldb_keymetadata_t  { leveldb::KeyMetaData       rep; };
```
paulplace, line 18: This declaration breaks stated convention #1 in c.h that the objects passed to the C API are opaque struct pointers.

Perhaps we don't really care about maintaining this characteristic of the C API. Whether we do or not, should probably add a comment here stating that this declaration must match the corresponding declaration in c.cc.

bashopatricia:
@matthewvon this is the open source version we discussed Friday, correct?  
create jira issue

bashopatricia:
create jira issue

matthewvon:
This will be the open source expiry.  It is still configured for EE since we were coding before product management decided.

db/c_test.c
```diff
@@ -112,6 +119,117 @@ static void CheckDel(void* ptr, const char* k, size_t klen) {
...
+static struct CheckPut2Data gCheckPut2ItrData[]=
+{
+    {"bar","b",leveldb::kTypeValue,0},
+    {"box","c",leveldb::kTypeValue,0},
+    {"bar","",leveldb::kTypeDeletion,0},
+    {"mom","texas",leveldb::kTypeValueWriteTime,0},
+    {"dad","poland",leveldb::kTypeValueExplicitExpiry,22446688}
+  };
+
+static void CheckPut2(void* ptr,
```
paulplace, line 203: CheckPut() and CheckDel() are no longer referenced and can be deleted.

db/c_test.c
```diff
@@ -329,6 +480,70 @@ int main(int argc, char** argv) {
...
+                    test->m_Type, test->m_Expiry);
+      }   // for
+
+      // close and open to force memory table into .sst upon open
+      leveldb_close(db);
+      leveldb_options_set_error_if_exists(options, 0);
+      db = leveldb_open(options, dbname, &err);
+      CheckNoError(err);
+
+      // now testing get from a level-0 .sst file
```
paulplace, line 529: Would be good if we could force compactions and check other levels, too. Not sure if this is feasible.

db/c_test.c
```diff
@@ -329,6 +480,70 @@ int main(int argc, char** argv) {
...
+          struct CheckPut2Data * test;
+
+          test=&gCheckPut2Data[loop];
+          klen=strlen(test->m_Key);
+          vlen=strlen(test->m_Value);
+
+          CheckGet2(db, roptions, test->m_Key, test->m_Value,
+                    test->m_Type, test->m_Expiry);
+      }   // for
+  }
```
paulplace, line 543: Do we have any tests where the key has expired? Perhaps by setting a short expiry window and sleeping between the put and get calls.

db/db_impl.cc
```diff
@@ -932,7 +944,23 @@ void DBImpl::BackgroundCall2(
...
+            s = BackgroundExpiry(Compact);
+            break;
+
+        default:
+            assert(0);
+            break;
+    }   // switch
+
     if (!s.ok() && !shutting_down_.Acquire_Load()) {
       // Wait a little bit before retrying background compaction in
```
paulplace, line 965: Should this wait be after any type of compaction or just a background compaction? If background only, then the condition needs to be adjusted; if background or expiry, then the logging statement below should be adjusted to indicate the type of compaction that was attempted.

db/db_iter.cc
```diff
@@ -70,6 +73,23 @@ class DBIter: public Iterator {
...
+  // REQUIRES: Valid() and forward iteration
+  //  (reverse iteration is possible, just needs code)
+  virtual KeyMetaData & keymetadata() const
+  {
+    assert(valid_ && kForward==direction_);
+    if (kForward==direction_)
+    {
+      ParsedInternalKey parsed;
+      ParseInternalKey(iter_->key(), &parsed);
+      keymetadata_.m_Type=parsed.type;
```
paulplace, line 86: Rather than setting individual data members of the KeyMetaData struct here, I suggest adding a setter method in KeyMetaData and calling it here. That way, if the members in KeyMetaData change, you only need to change the signature of the setter and the compiler will tell you any place that you call it with the incorrect member values.

db/dbformat.h
```diff
@@ -272,6 +278,30 @@ class LookupKey {
   Slice user_key() const
   { return Slice(kstart_, end_ - kstart_ - KeySuffixSize(internal_key())); }
 
+  // did requestor have metadata object?
+  bool WantsKeyMetaData() const {return(NULL!=meta_);};
+
+  void SetKeyMetaData(ValueType type, SequenceNumber seq, ExpiryTime expiry) const
+  {if (NULL!=meta_)
```
paulplace, line 285: Since this is a multi-line code block, should probably move the "if" to the next line. Same comment for the next method.

db/dbformat.cc
```diff
@@ -3,6 +3,7 @@
 // found in the LICENSE file. See the AUTHORS file for names of contributors.
 
 #include <stdio.h>
+//#include "leveldb/expiry.h"
```
paulplace, line 6: If this inclusion is not needed, then remove this line.

db/dbformat.cc
```diff
@@ -147,7 +148,8 @@ bool InternalFilterPolicy::KeyMayMatch(const Slice& key, const Slice& f) const {
   return user_policy_->KeyMayMatch(ExtractUserKey(key), f);
 }
 
-LookupKey::LookupKey(const Slice& user_key, SequenceNumber s) {
+  LookupKey::LookupKey(const Slice& user_key, SequenceNumber s, KeyMetaData * meta) {
```
paulplace, line 151: This line should not be indented

db/dbformat.cc
```diff
@@ -183,12 +186,19 @@ KeyRetirement::KeyRetirement(
 }   // KeyRetirement::KeyRetirement
 
 
+KeyRetirement::~KeyRetirement()
+{
+    if (0!=expired)
+        gPerfCounters->Add(ePerfExpiredKeys, expired);
+}   // KeyRetirement::~KeyRetirement
```
paulplace, line 193: What about the `dropped` data member? Should it be checked here, too?

db/memtable.h
```diff
@@ -72,7 +76,7 @@ class MemTable {
     int operator()(const char* a, const char* b) const;
   };
   friend class MemTableIterator;
-  friend class MemTableBackwardIterator;
+  friend class MemTableBackwardIterator; // does not exist
```
paulplace, line 79: So, should we get rid of this declaration?

db/memtable.cc
```diff
@@ -81,7 +84,8 @@ Iterator* MemTable::NewIterator() {
 
 void MemTable::Add(SequenceNumber s, ValueType type,
                    const Slice& key,
-                   const Slice& value) {
+                   const Slice& value,
+                   const ExpiryTime & expiry) {
   // Format of an entry is concatenation of:
```
paulplace, line 89: Would be good to update this comment to match what is actually put into the buffer.

db/version_edit.h
```diff
@@ -119,8 +121,9 @@ class VersionEdit {
   void DeleteFile(int level, uint64_t file) {
     deleted_files_.insert(std::make_pair(level, file));
   }
+  size_t DeletedFileCount() const {return(deleted_files_.size());};
 
-  void EncodeTo(std::string* dst, bool format2=false) const;
+  void EncodeTo(std::string* dst, bool format2=true) const;
```
paulplace, line 128: Anticipating the possible need in the future to extend the serialized format, it might be a good idea to change the format indicator from a bool to an enumeration.

db/version_set.cc
```diff
@@ -1792,13 +1856,15 @@ Compaction::Compaction(int level)
...
       grandparent_index_(0),
       seen_key_(false),
       overlapped_bytes_(0),
       tot_user_data_(0), tot_index_keys_(0),
       avg_value_size_(0), avg_key_size_(0), avg_block_size_(0),
       compressible_(true),
-      stats_done_(false)
+      stats_done_(false),
+      no_move_(false)
   {
```
paulplace, line 1868: This opening brace should not be indented.

db/write_batch.cc
```diff
@@ -53,13 +55,13 @@ Status WriteBatch::Iterate(Handler* handler) const {
...
     found++;
-    char tag = input[0];
+    ValueType tag = (ValueType)input[0];
     input.remove_prefix(1);
     switch (tag) {
       case kTypeValue:
         if (GetLengthPrefixedSlice(&input, &key) &&
             GetLengthPrefixedSlice(&input, &value)) {
-          handler->Put(key, value);
+            handler->Put(key, value, kTypeValue, 0);
```
paulplace, line 64: Something feels a bit "off" to call `Handler::Put()` with values (`kTypeValue` and `0`) when the method is declared to take const references. The compiler allows it, which means it is likely putting the values in temporary anonymous variables and passing references to them, but it still looks funny. Not really suggesting anything here, just observing.

db/write_batch.cc
```diff
@@ -108,26 +110,20 @@ void WriteBatchInternal::SetSequence(WriteBatch* b, SequenceNumber seq) {
...
+void WriteBatch::Put(const Slice& key, const Slice& value, const KeyMetaData * meta) {
+  KeyMetaData local_meta;
   WriteBatchInternal::SetCount(this, WriteBatchInternal::Count(this) + 1);
-  rep_.push_back(static_cast<char>(kTypeValueExplicitExpiry));
+  if (NULL!=meta)
+      local_meta=*meta;
+  rep_.push_back(static_cast<char>(local_meta.m_Type));
   PutLengthPrefixedSlice(&rep_, key);
-  PutVarint64(&rep_, expiry);
+  if (kTypeValueExplicitExpiry==local_meta.m_Type
```
paulplace, line 120: Might be a bit cleaner to add an `IsExpiryType()` method to the `KeyMetaData` struct and call it here (and possibly elsewhere) instead of doing explicit comparisons against `m_Type`.

db/write_batch_test.cc
```diff
@@ -100,20 +100,26 @@ TEST(WriteBatchTest, Multiple) {
 
 TEST(WriteBatchTest, MultipleExpiry) {
   WriteBatch batch;
+  KeyMetaData meta;
   batch.Put(Slice("Mary"), Slice("Lamb"));
-  batch.PutExplicitExpiry(Slice("Adam"), Slice("Ant"), 2347);
+  meta.m_Type=kTypeValueExplicitExpiry;
+  meta.m_Expiry=2347;
+  batch.Put(Slice("Adam"), Slice("Ant"), &meta);
+  //batch.PutExplicitExpiry(Slice("Adam"), Slice("Ant"), 2347);
```
paulplace, line 108: This line can be deleted since `PutExplicitExpiry()` was deleted.

include/leveldb/expiry.h
```diff
@@ -0,0 +1,98 @@
...
+class SstCounters;
+class Version;
+class VersionEdit;
+struct FileMetaData;
+
+class ExpiryModule : public RefObjectBase
+{
+public:
+    ExpiryModule() {};
+    ~ExpiryModule() {};
```
paulplace, line 45: The dtor is implicitly virtual since the dtor in `RefObjectBase` base class is declared as virtual, but I still think it's helpful to declare it explicitly as virtual in derived classes.

db/write_batch.cc
```diff
@@ -142,23 +138,33 @@ class MemTableInserter : public WriteBatch::Handler {
...
+  MemTableInserter() : mem_(NULL), options_(NULL) {};
+
+  virtual void Put(const Slice& key, const Slice& value, const ValueType &type, const ExpiryTime &expiry) {
+    ValueType type_use(type);
+    ExpiryTime expiry_use(expiry);
 
-  virtual void Put(const Slice& key, const Slice& value) {
-    mem_->Add(sequence_, kTypeValue, key, value);
+    if (NULL!=options_ && NULL!=options_->expiry_module.get())
+        options_->expiry_module->MemTableInserterCallback(key, value, type_use, expiry_use);
```
paulplace, line 150: Should the return value of `MemTableInserterCallback()` be checked here?

table/table_builder.cc
```diff
@@ -138,6 +139,17 @@ void TableBuilder::Add(const Slice& key, const Slice& value) {
...
 
+  // again ignore short keys, save high sequence number for abbreviated repair
+  if (8 <= key.size()
+      && r->sst_counters.Value(eSstCountSequence)<ExtractSequenceNumber(key))
+      r->sst_counters.Set(eSstCountSequence,ExtractSequenceNumber(key));
+
+  // statistics if an expiry key
+  if (NULL!=r->options.expiry_module.get())
+  {
+      r->options.expiry_module->TableBuilderCallback(key, r->sst_counters);
```
paulplace, line 150: Should the return value of `TableBuilderCallback()` be checked? Log an error message?

include/leveldb/expiry.h
```diff
@@ -0,0 +1,98 @@
...
+
+    // Print expiry options to LOG file
+    virtual void Dump(Logger * log) const
+    {Log(log,"                        Expiry: (none)");};
+
+    // db/write_batch.cc MemTableInserter::Put() calls this.
+    // returns false on internal error
+    virtual bool MemTableInserterCallback(
+        const Slice & Key,   // input: user's key about to be written
```
paulplace, line 54: Would be nice to line up the comments as is done below for `CompactionFinalizeCallback()`.

table/table.cc
```diff
@@ -111,46 +109,49 @@ void Table::ReadMeta(const Footer& footer) {
...
+                  ReadFilter(iter->value(), policy);
+                  gPerfCounters->Inc(ePerfBlockFilterRead);
+                  found=true;
+              }   // if
+          }   //if
+      } while(!found && NULL!=policy);
+  }   // if
+
+  // always read counters
   key="stats.sst1";
```
paulplace, line 155: Is this a key in the user-visible key space? If so, I worry that it is not unique enough to avoid collision with a user key.

The same concern applies to the "filter." key above.

paulplace:
+1 9e47fbb

Transition from EE to OS looks good. All unit tests pass

PR #188, opened 2016-02-29
Correct DB::Open() to use running parameters, not user parameters.

matthewvon:
Make sure DB::Open() uses working parameters, not user parameters.  Fixes data loss in tiered storage scenario.

Detailed discussion is here:
  https://github.com/basho/leveldb/wiki/mv-tiered-recovery-log


paulplace:
+1 11e84f5

PR #192, opened 2016-04-15
Fix then disable timed grooming

matthewvon:
This branch fixes an oversight in the original timed grooming implementation.  This branch also disables timed grooming until its block cache impact is mitigated.

Details are here:  https://github.com/basho/leveldb/wiki/mv-timed-grooming2

### DIFF
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -1124,6 +1124,9 @@ VersionSet::Finalize(Version* v)
             else
                 elapsed_micros=0;
 
+            // reevaluating timed grooming ... seems to crush caching
+            elapsed_micros=0;
+
             // which grooming trigger point?  based upon how long
             //  since last compaction on this level
             //   - less than 10 minutes?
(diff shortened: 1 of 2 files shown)


db/version_set.h
```diff
@@ -316,7 +316,10 @@ class VersionSet {
   void SetCompactionDone(int level, uint64_t Now)
   {   m_CompactionStatus[level].m_Running=false;
       m_CompactionStatus[level].m_Submitted=false;
-      m_CompactionStatus[level].m_LastCompaction=Now; }
+      m_CompactionStatus[level].m_LastCompaction=Now; 
+      if ((level+1)<config::kNumLevels)
```
paulplace, line 323: Would be good to add a comment here indicating why you care about level+1

db/version_set.cc
```diff
@@ -1124,6 +1124,9 @@ VersionSet::Finalize(Version* v)
             else
                 elapsed_micros=0;
 
+            // reevaluating timed grooming ... seems to crush caching
+            elapsed_micros=0;
```
paulplace, line 1130: Would be good to have a comment here stating that this explicitly and intentionally disables the timed grooming feature

paulplace:
+1 67a99dd

PR #194, opened 2016-06-03
Switch from <cstdatomic> to <atomic>

sharkcz:
Switch from <cstdatomic> to <atomic>. The former never made it into the
standard and doesn't exist in modern gcc versions at all.  The later contains
everything that leveldb was using from the former.

Backported from Google leveldb 1.18
https://github.com/google/leveldb/commit/803d69203a62faf50f1b77897310a3a1fcae712b


matthewvon:
Thank you.  Will be a couple of weeks before I can circle back to this for proper review and test cycle.

matthewvon:
Basho has closed down.  I do not know of anyone that has access to merge this PR.  Just letting you know.

lemenkov:
This fixes #230. Since the project restarted its activity, then maybe we should merge it to develop?

PR #200, opened 2016-07-12
Replace QueueThread class with simple mutex

matthewvon:
The QueueThread class in util/hot_threads.cc exists to correct a potential race condition.  This is a simpler fix to same race condition that utilizes an existing mutex.  Details here:

```
https://github.com/basho/leveldb/wiki/mv-no-semaphore
```


paulplace:
+1 175cb2f

PR #201, opened 2016-07-15
add sed command that removes cuttlefish unit test disable tags.

matthewvon:
This adds one sed line to the script ppackager.sh pushes to a buildbot.  The sed line removes comment markers that disable the cuttlefish dependency specification and cuttlefish unit tests.


paulplace:
+1 410caea

PR #202, opened 2016-08-03
Mv hot backup [JIRA: RIAK-2784]

matthewvon:
Hot Backup allows an external trigger file, /etc/riak/hot_backup, to cause all open leveldb databases on a server to create a backup image, live.  The leveldb process does not require the user to stop or pause data operations.  Hot backup will maintain up to five backup snapshots in a rolling fashion if the user does not wish to copy and erase after each backup.

Each backup snapshot is "ready to run".  Previous Basho and 3rd party solutions required a database "repair" or similar lengthy process to guarantee state before using their backup.

Usage and code details are here:  https://github.com/basho/leveldb/wiki/mv-hot-backup


util/hot_backup_test.cc
```diff
@@ -0,0 +1,600 @@
...
+
+    // preclean
+    DestroyDB(backup_path, options);
+    rmdir(backup_path.c_str());
+    DestroyDB(m_DBName, options);
+    rmdir(m_DBName.c_str());
+
+    // need a live database for this test.
+    options.create_if_missing=true;
+    options.error_if_exists;
```
paulplace: This line does nothing except generate a build warning. Should it be set to true? If not, please remove.

The same comment applies to lines 336 and 511.

matthewvon: Oops.  Where are you building?  OS X?  Want to duplicate the warning.

paulplace: Yes, on OS X. Setting up to build on Debian as I write this.

matthewvon: Not arguing the bug.  Just want to duplicate.  Bug is pretty damn obvious now that you point it out.

hqqq
```diff
@@ -0,0 +1,76 @@
+diff --git a/util/hot_backup.cc b/util/hot_backup.cc
```
paulplace: This file looks like something that was checked in unintentionally.

matthewvon: now gone.  no clue how it was created.  but pretty clear on who let the random file into the repository.  me.  idiot.

bashopatricia:
create jira issue

Basho-JIRA:
Tagged this as Riak 2.3.0 so we evaluate when the Riak team starts looking at the next release [~angrycub] 

util/hot_backup.h
```diff
@@ -0,0 +1,127 @@
...
+// Called every 60 seconds to test for external hot backup trigger
+//   (initiates backup if trigger seen)
+void CheckHotBackupTrigger();
+
+
+/**
+ * Singleton HotBackup object that unit tests can override.
+ */
+
+extern class HotBackup * gHotBackup;
```
paulplace: Rather than exposing the HotBackup singleton as a global static variable, I recommend making it a private data member of the HotBackup class. Then remove the "unit test support" routines from the public interface of the HotBackup class and add a protected constructor that takes a pointer to the HotBackup object to use as the singleton. For unit tests, have the HotBackupTester class pass its "this" pointer in the base class constructor to set the singleton.

UPDATE: would also need to make CheckHotBackupTrigger() a public static method in the HotBackup class so that it could access the private static gHotBackup data member.

As a general rule, I would like us to move away from having "test only" data members and methods publicly exposed by a class.

util/hot_backup.h
```diff
@@ -0,0 +1,127 @@
...
+
+#include "util/thread_tasks.h"
+
+namespace leveldb
+{
+
+namespace config {
+
+// how many hot backup directories:  backup, backup.1, ... backup.(kNumBackups-1)
+static const int kNumBackups = 5;
```
paulplace, line 34: I can easily imagine the next version of this feature requiring that the number of backups be configurable (and possibly the path of the trigger file, too).

Could easily prepare for that possibility by making these data members of the HotBackup class. They could be initialized in the constructor with the current default values, but then provide get/set methods that allow later code to adjust these settings based on a riak configuration setting.

UPDATE: This approach would also making testing these configuration settings easier.

util/hot_backup.cc
```diff
@@ -0,0 +1,654 @@
...
+ * Look for external trigger
+ */
+bool
+HotBackup::IsTriggerSet()
+{
+    bool ret_flag;
+    int ret_val;
+
+    //
+    // This is code polls for existance of /etc/riak/hot_backup
```
paulplace: "existence" is mispelled

util/hot_backup.cc
```diff
@@ -0,0 +1,654 @@
...
+
+
+/**
+ * Let external trigger know that hot backup is complete.  Put
+ *  message to syslog if unable to delete trigger file
+ */
+void
+HotBackup::ResetTrigger()
+{
+    bool ret_flag;
```
paulplace: ret_flag variable is not used

util/hot_backup.cc
```diff
@@ -0,0 +1,654 @@
...
+
+    // hold mutex while reviewing database state
+    //  (note:  never good to call Log() while holding mutex
+    //   but these are unlikely error cases)
+    {
+        // get mutex,
+        MutexLock l(&mutex_);
+
+        // set backup flag && see if
+        assert(false==hotbackup_pending_);
```
paulplace, line 189: Should this assert be removed (or at least commented out)? Didn't we recently have a weird "riak dies every so often" bug caused by an assert that eleveldb fails to disable in a release build?

util/hot_backup.cc
```diff
@@ -0,0 +1,654 @@
...
+    ret_val=unlink(GetTriggerPath());
+    if (0==ret_val)
+    {
+        // release last counter, enabling future backups
+        HotBackupFinished();
+    }   // if
+    else
+    {
+        // one count left, effectively disabling trigger forever (or next program restart)
+        Log(NULL, "leveldb HotBackup unable to delete trigger file %s", GetTriggerPath());
```
paulplace: Would be good to log errno here. Could even call strerror() if you want to get fancy.

util/hot_backup.cc
```diff
@@ -0,0 +1,654 @@
...
+HotBackup::ResetTrigger()
+{
+    bool ret_flag;
+    int ret_val;
+
+    ret_val=unlink(GetTriggerPath());
+    if (0==ret_val)
+    {
+        // release last counter, enabling future backups
+        HotBackupFinished();
```
paulplace, line 116: HotBackupFinished() calls this method when m_JobsPending goes to 1. I'm unclear why we need that circularity.

include/leveldb/perf_count.h
```diff
@@ -240,6 +240,9 @@ enum PerformanceCountersEnum
     ePerfExpiredKeys=94,      //!< key physically removed because it expired
     ePerfExpiredFiles=95,     //!< entire file removed because all keys expired
 
+    ePerfSyslogWrite=96,      //!< logged message to syslog
+    ePerfBackupStarted=97,    //!< hot backup initiated
```
paulplace, line 244: Might be useful to also count backups completed.

UPDATE: and backups that had an error

util/hot_backup.cc
```diff
@@ -0,0 +1,654 @@
+// -------------------------------------------------------------------
+//
+// hot_backup.cc
+//
+// Copyright (c) 2011-2016 Basho Technologies, Inc. All Rights Reserved.
```
paulplace: For new files, the copyright date should not be a range but rather it should be just the current year, since this file (and the corresponding .h file) were first created this year.

util/hot_backup.cc
```diff
@@ -0,0 +1,654 @@
...
+                Log(LiveOptions.info_log, "HotBackup failed while renaming %s directory",
+                    src_path.c_str());
+            }   // else
+        }   // for
+
+    }   // if
+    else
+    {
+        good=false;
+        Log(LiveOptions.info_log, "HotBackup failed while removing %s directory",
```
paulplace: This message will be misleading if the removal failed on the slow tier. Does the status object have any more info that could be logged?

util/hot_backup.cc
```diff
@@ -0,0 +1,654 @@
...
+                Log(GetLogger(), "HotBackup LOG read failed at %ld (%s)",
+                    remaining, s.ToString().c_str());
+        }   // else
+
+        remaining-=count;
+    }   // while
+
+    // close files and go home
+    good=s.ok();
+    delete src;
```
paulplace, line 646: I recommend making src and dst std::auto_ptr<> objects so that their lifetimes are automatically managed.

util/hot_backup.cc
```diff
@@ -0,0 +1,654 @@
...
+    local_options=options_;
+    SetBackupPaths(local_options, 0);
+
+    src_name=InfoLogFileName(options_.tiered_fast_prefix);
+    dst_name=InfoLogFileName(local_options.tiered_fast_prefix);
+
+    s=options_.env->NewSequentialFile(src_name, &src);
+    if (s.ok())
+    {
+        s=options_.env->NewWritableFile(dst_name, &dst, 4096);
```
paulplace: 4096 is hard-coded several times in this method; should this be a method-level, or possibly class-level, constant? Could it need to change based on the platform (in which case, it should probably be a class property that can be changed)?

util/hot_backup.cc
```diff
@@ -0,0 +1,654 @@
...
+    // purge imm or current write buffer
+    good=good && m_DBImpl.PurgeWriteBuffer();
+
+    // Gather random data items for MANIFEST and
+    //  other supporting files.
+    if (good)
+    {
+        // get LOG file size
+        if (NULL!=m_DBImpl.GetLogger())
+            log_position=m_DBImpl.GetLogger()->LogSize();
```
paulplace, line 247: For the default Logger (PosixLogger), the LogSize() method calls ftell(), which returns -1 on error, so log_position should be checked before using it below. Either that, or change the conditional below from "0!=log_position" to "0<log_position", if doing everything else still makes sense.

util/hot_backup.cc
```diff
@@ -0,0 +1,654 @@
...
+    {
+        log::Writer log(file);
+        std::string record;
+        edit.EncodeTo(&record);
+        status = log.AddRecord(record);
+
+        if (status.ok())
+            status = file->Close();
+
+        delete file;
```
paulplace, line 488: I recommend making file a std::auto_ptr<> and changing this delete statement to auto_ptr<>::reset().

util/hot_backup.cc
```diff
@@ -0,0 +1,654 @@
...
+        {
+            src_path=TableFileName(options_, (*it)->number, level);
+            dst_path=TableFileName(BackupOptions, (*it)->number, level);
+
+            // this would be cleaner if link operation added to Env class heirarchy
+            ret_val=link(src_path.c_str(), dst_path.c_str());
+            if (0!=ret_val)
+            {
+                good=false;
+                Log(GetLogger(), "HotBackup failed linking (return val %d) %s",
```
paulplace: Would be good to log errno here, too, since ret_val is most likely -1.

util/hot_backup.cc
```diff
@@ -0,0 +1,654 @@
...
+        Log(GetLogger(), "HotBackup failed in PurgeWriteBuffer");
+
+    return(good);
+
+}   // DBImpl::PurgeWriteBuffer
+
+
+/**
+ * Need an independent MANIFEST and CURRENT file within
+ * hot backup directory.  This creates them.
```
paulplace, line 422: Recommend adding to this comment that this method also creates the file links. Could even change this method name to better indicate the tasks it performs; maybe something like WriteBackupManifestAndLinkFiles().

util/hot_backup.cc
```diff
@@ -0,0 +1,654 @@
...
+{
+    assert(0<EndPos);
+
+    bool good(true);
+    long remaining;
+    Options local_options;
+    SequentialFile * src(NULL);
+    WritableFile * dst(NULL);
+    Status s;
+    std::string src_name, dst_name, buffer;
```
paulplace, line 590: I recommend changing the buffer variable to be of type std::vector<char>, since that better reflects how you're using it.

util/hot_backup.cc
```diff
@@ -0,0 +1,654 @@
...
+                Log(GetLogger(), "HotBackup LOG write failed at %ld (%s)",
+                    remaining, s.ToString().c_str());
+        }   // if
+        else
+        {
+                Log(GetLogger(), "HotBackup LOG read failed at %ld (%s)",
+                    remaining, s.ToString().c_str());
+        }   // else
+
+        remaining-=count;
```
paulplace: Shouldn't this be subtracting data_read from remaining instead of count?

util/hot_backup.cc
```diff
@@ -0,0 +1,654 @@
...
+        count=(4096<remaining ? 4096 : remaining);
+        s=src->Read(count, &data_read, (char *)buffer.data());
+
+        if (s.ok())
+        {
+            s=dst->Append(data_read);
+
+            if (!s.ok())
+                Log(GetLogger(), "HotBackup LOG write failed at %ld (%s)",
+                    remaining, s.ToString().c_str());
```
paulplace: This log message will be a bit misleading since you're logging the bytes remaining value but the log message implies that you're logging the offset of the write. The same applies to the Log() message below.

If you want to log the offset where you're currently reading/writing (which makes sense), you'll need to add another variable to track the current read/write offset.

util/hot_backup.cc
matthewvon: Good eye!  Wrong repository.  Google's Makefile knows to disable assert() in production builds and leave active in debug builds.  Their asserts have saved my ass a couple of times, especially related to keys.  (Yes, I reworked their Makefile but left that important feature!)

Yes, eleveldb dies on the one assert() I left in that code.  But lately it has proved better to have that assert() fire than wait for a random, likely double free, bug.  That said, I do not recommend leaving live asserts as a defense mechanism.  Someday I will rewrite how the C++ code builds in eleveldb.

util/hot_backup.cc
matthewvon: I read somewhere that std::auto_ptr<> is being deprecated.  Do you happen to know its replacement?

That said, I will use the auto_ptr since our build environments are way, way behind the standards.  Question is a personal curiosity.

util/hot_backup_test.cc
```diff
@@ -0,0 +1,600 @@
...
+{
+    char * dup_path, *path;
+    int ret_val;
+    bool ret_flag;
+    uint64_t perf_before, perf_after;
+
+    perf_before=gPerfCounters->Value(ePerfSyslogWrite);
+
+    // cleanup anything existing, likely fails
+    ///  hmm, should there be a way to move this trigger
```
paulplace: It looks like the HotBackupTester class sets a trigger file path in a temp directory, so I'm guessing this is an old comment (that can be removed).

util/hot_backup.cc
matthewvon: My understanding is that std::string is guaranteed to be contiguous, std::vector is highly likely to be contiguous ... but not guaranteed.  Do you have a clear source on implementations?

util/hot_backup_test.cc
```diff
@@ -0,0 +1,600 @@
...
+    //  bypass test if it does not ... Travis CI and
+    //  other users might not be able to access /etc/riak
+    dup_path=strdup(GetTriggerPath());
+    path=dirname(dup_path);
+    ret_val=access(path, R_OK | W_OK);
+    ASSERT_TRUE(-1!=ret_val);
+
+    // is a trigger seen (hope not)
+    ret_flag=HotBackup::IsTriggerSet();
+    ASSERT_TRUE(!ret_flag);
```
paulplace: I recommend extending util/testharness.h, adding ASSERT_FALSE, and calling that here (seems like it should have already been there).

util/hot_backup_test.cc
```diff
@@ -0,0 +1,600 @@
...
+    ASSERT_TRUE(ret_flag);
+
+    // release one, trigger goes away
+    HotBackup::HotBackupFinished();
+    ret_flag=HotBackup::IsTriggerSet();
+    ASSERT_TRUE(!ret_flag);
+
+    // did our simulation create a syslog entry?
+    //  (bonus if you manually check /var/log/syslog for actual entries)
+    perf_after=gPerfCounters->Value(ePerfSyslogWrite) - perf_before;
```
paulplace: perf_after could be more accurately named perf_diff

util/hot_backup_test.cc
```diff
@@ -0,0 +1,600 @@
...
+    ASSERT_TRUE( 1==perf_after );
+
+    // clean up second count.
+    HotBackupFinished();
```
paulplace: If I'm reading the code correctly (and I'll be the first to admit that I may not be doing so), it looks like this call is unnecessary. Here's what I _think_ happens: the first call to HotBackup::HotBackupFinished() (above) decrements m_JobsPending to 1 and calls ResetTrigger(), which deletes the trigger file and calls HotBackupFinished(). The second (recursive) call to HotBackupFinished() then decrements m_JobsPending to 0.

With all that said, I don't think this call to HotBackupFinished() causes a problem, but its associated comment is misleading (if my analysis above is correct).

util/hot_backup.cc
paulplace, line 189: Ah, so the copy of leveldb embedded in eleveldb still does the correct
thing with asserts? That's good news. I thought we lost all "no asserts in
non-debug builds" behavior when building eleveldb and its dependencies.

util/hot_backup.cc
paulplace, line 488: C++11 added std::unique_ptr<>, which supersedes std::auto_ptr<>, and adds
some additional goodness (such as properly handling arrays).

I tried numerous times to move our toolset forward so that I could start
using unique_ptr<>, but to no avail. Yes, you're stuck with auto_ptr<> for
now.

util/hot_backup.cc
matthewvon: Decided to review an eleveldb build log before making an incorrect statement:

mac-mini.bos1 osx10.8   cc -g -O2 -DERTS_FRMPTR -Wall -O3 -fPIC -I /Users/buildbot/matthewv/eleveldb/c_src/system/include -I. -I./
include -fno-builtin-memcmp -DOS_MACOSX -DLEVELDB_PLATFORM_POSIX  -DLEVELDB_VSN="2.0.27" -O2 -g -DNDEBUG     -fPIC -MM -E -MT util
/bloom2.d -MT util/bloom2.o -MF util/bloom2.d util/bloom2.cc

The parameter "-DNDEBUG" is what disables asserts.  It is active in the eleveldb initiated builds.  My previous statement is therefore correct (and verified).

util/hot_backup.cc
paulplace: Holy crap! I have never seen an implementation of std::vector<> that didn't allocate a contiguous memory block for the underlying storage, but apparently the C++98 spec doesn't preclude doing so.

I just found a discussion about this on stackoverflow (http://stackoverflow.com/questions/849168/are-stdvector-elements-guaranteed-to-be-contiguous). Apparently this omission was corrected in C++98-TC1, and the new C++11 standard makes this requirement explicit.

Without looking at the implementations of some of our "problem child" compilers (e.g., FreeBSD and Solaris), I can't say whether or not they do the right thing with vector allocation, so sticking with your std::string approach seems wise for now.

util/hot_backup_test.cc
```diff
@@ -0,0 +1,600 @@
...
+    ASSERT_OK(s);
+    s=db->Get(read_ops, "14", &out_buffer);
+    ASSERT_OK(s);
+    s=db->Get(read_ops, "25", &out_buffer);
+    ASSERT_OK(s);
+    s=db->Get(read_ops, "26.2", &out_buffer);
+    ASSERT_OK(s);
+    // test fail case
+    s=db->Get(read_ops, "26.1", &out_buffer);
+    ASSERT_TRUE(!s.ok());
```
paulplace: Can use ASSERT_NOTOK() here, if you're so inclined.

util/hot_backup.h
```diff
@@ -0,0 +1,127 @@
...
+
+static const char * kTriggerFileName="/etc/riak/hot_backup";
```
paulplace: I recommend changing the default name of the trigger file to "/etc/riak/leveldb_hot_backup" to more accurately reflect its intent in the riak ecosystem (i.e., that file triggers a leveldb hot backup, not a hot backup of something else). If you want to decouple leveldb from riak, could use "/etc/leveldb/hot_backup".

util/hot_backup.cc
```diff
@@ -0,0 +1,654 @@
...
+    const Options & LiveOptions)
+{
+    Options local_options;
+    bool good;
+    Status status;
+
+    assert(0 < config::kNumBackups);
+    good=true;
+
+    // Destroy highest possible directory if it exists
```
paulplace, line 291: A suggestion for a future version of this feature is to defer deleting the highest possible directory until you've created the new backup directory (just so that you don't lose data if you cannot perform the backup).

util/hot_backup.cc
```diff
@@ -0,0 +1,654 @@
...
+        // retrieve current version and increment refs_ (requires mutex)
+        //  (includes code that creates hard links to .sst files)
+        good=m_DBImpl.WriteBackupManifest();
+
+        // last step is to replicate as much of LOG file from time of
+        //  backup (close, but not exact)
+        //  Assumes that 0 for log_position implies user has a custom
+        //  Logger class that this routine will fail to copy.
+        if (good && 0!=log_position)
+            good=m_DBImpl.CopyLOGSegment(log_position);
```
paulplace, line 260: Something that's not clear to me is how you ensure that you don't get a corrupted log file. That is, when you call LogSize() above, do you know for sure that another thread isn't in the middle of writing to the log file?

util/hot_backup.cc
```diff
@@ -0,0 +1,654 @@
...
+    bool create_backup_event(false);
+
+    // hold mutex while reviewing database state
+    //  (note:  never good to call Log() while holding mutex
+    //   but these are unlikely error cases)
+    {
+        // get mutex,
+        MutexLock l(&mutex_);
+
+        // set backup flag && see if
```
paulplace: This comment looks incomplete; either that, or I simply don't understand it

util/hot_backup_test.cc
```diff
@@ -0,0 +1,600 @@
...
+    DestroyDB(backup_path, options);
+    rmdir(backup_path.c_str());
+    DestroyDB(db_path, options);
+    rmdir(db_path.c_str());
+
+
+}   // DoubleDBTest
+
+}   // namespace leveldb
+
```
paulplace, line 593: Overall, this is a very good set of unit tests.

Another test that would be good to have is one where we test running a bunch of hot backup jobs while another thread is adding a bunch of records and starting compactions.

util/hot_backup.h
matthewvon: Done.

util/hot_backup.h
matthewvon: Going to wait on this.  The "config" namespace is established in db/dbformat.h for several values that could also become configurable some day ... but have not.  Going to leave this static until there is a customer requirement.

util/hot_backup.cc
matthewvon: corrected.

util/hot_backup.cc
matthewvon: Done.

util/hot_backup.cc
matthewvon: Done.  Passed on strerror().  It is not thread safe and not willing to put in the time to code and test strerror_r today.

util/hot_backup.cc
matthewvon: The circularity addresses a potential race condition.  It makes sure the trigger file is erased before the entire hot backup could be started again.  Half of the explanation is attempted in CheckHotBackupTrigger().  CheckHotBackupTrigger() happens on throttle's thread.  Each leveldb database backups up on a worker thread.  It would be possible for m_JobsPending to decrement to 0 on worker thread, throttle thread see it zero and start a new backup, then ResetTrigger() to erase the file on worker thread ... confusing state.

include/leveldb/perf_count.h
matthewvon: only need one or the other ... and Matthew is only known person to look at these values.  Adding error.

util/hot_backup.cc
matthewvon: done

util/hot_backup.cc
matthewvon: Updated error messages.

util/hot_backup.cc
matthewvon: Unified to match production code:  gMapSize.

util/hot_backup.cc
matthewvon: Adjusted LogSize() to return zero on error.  Good catch.

util/hot_backup.cc
matthewvon: done

util/hot_backup.cc
matthewvon: Gonna pass on this one.

util/hot_backup.cc
matthewvon: Ouch!  nasty bug.  fixed.

util/hot_backup_test.cc
matthewvon: done

util/hot_backup_test.cc
matthewvon: Done.

util/hot_backup_test.cc
matthewvon: fixed

util/hot_backup.cc
matthewvon: Done

util/hot_backup_test.cc
matthewvon: done

util/hot_backup.h
matthewvon: Going to pass this suggestion to Charlie V.

util/hot_backup.cc
matthewvon: LOG file is best efforts basis.  It is not essential for reuse of the data.  In fact, it gets moved to LOG.old and a new LOG file installed upon start of leveldb.  The underlying question is how do we know when ftell() gets updated:  before a printf to LOG, after, or some unknown time.  I am not going to try and solve that.

util/hot_backup.cc
matthewvon: Comments corrected.

util/hot_backup.cc
matthewvon: My issue is that the open file calls require a pointer to the pointer.  I would need to have a simple pointer to pass into the open file calls.  Then shove it into the auto_ptr ... or rewrite the open file calls to take an auto_ptr.  Not in the mood for either at this time.  auto_ptr is better code, but feels like we would be taking a hammer to a screw.

util/hot_backup.cc
matthewvon: My issue is that the open file calls require a pointer to the pointer.  I would need to have a simple pointer to pass into the open file calls.  Then shove it into the auto_ptr ... or rewrite the open file calls to take an auto_ptr.  Not in the mood for either at this time.  auto_ptr is better code, but feels like we would be taking a hammer to a screw.

util/hot_backup_test.cc
matthewvon: done

util/hot_backup.cc
matthewvon: Hmm.  My first thought was in agreement.  Then I realized that we might have to recover the space from this backup before the new backup can succeed.  Hmm.  Likely throw this question out to Charlie.

util/hot_backup_test.cc
matthewvon: ok, but not today.

paulplace:
+1 9d16158

Basho-JIRA:
Hidden feature in Riak 2.2.0.  Limited beta.  Riak 2.3.0 is target general release.  

Basho-JIRA:
[~drohrer][~fdushin] I am probably the one who is just late on this, but Hot Backup was added as a hidden feature in Riak 2.2.0.  Limited beta to just a few customers and CliServ that [~angrycub] will drive/define.  Riak 2.3.0 is the target general release. Training will include information for CliServ and Sales so they know that this feature is not to be promoted.   

PR #203, opened 2016-08-19
Mv expiry iter bug

matthewvon:
DBImpl::NewIterator() was not setting expiry module when creating iterator object.

A detail discussion of this fix is found here:

https://github.com/basho/leveldb/wiki/mv-exiry-iter-bug

### DIFF
--- a/leveldb_os/expiry_os_test.cc
+++ b/leveldb_os/expiry_os_test.cc
@@ -22,6 +22,7 @@
 
 #include <limits.h>
 #include <algorithm>
+#include <memory>
 #include <string>
 
 #include "util/testharness.h"
@@ -791,6 +792,10 @@ public:
 
         m_Options.create_if_missing=true;
         m_Options.error_if_exists=false;
+
+        // Note: m_Options.expiry_module is a smart pointer.  It
+        //  owns the m_Expiry object and will automatically delete the
+        //  allocation.
         m_Expiry=new ExpTestModule;
         m_Options.expiry_module=m_Expiry;
 
@@ -1253,6 +1258,7 @@ TEST(ExpiryManifestTester, Overlap2)
 
     // let multiple threads complete
     /// sleep(1) required for Smart OS 1.8 buildbot
+    ///  then rased to sleep(2) for freebsd buildbot
     sleep(2);
     VerifyFiles(Overlap1, manifest_count, 5);
 
@@ -1372,6 +1378,10 @@ public:
 
         m_Options.create_if_missing=true;
         m_Options.error_if_exists=false;
+
+        // Note: m_Options.expiry_module is a smart pointer.  It
+        //  owns the m_Expiry object and will automatically delete the
+        //  allocation.
         m_Expiry=new leveldb::ExpiryModuleOS;
         m_Options.expiry_module=m_Expiry;
 
@@ -1385,13 +1395,6 @@ public:
         leveldb::DestroyDB(m_DBName, m_Options);
     };
 
-    bool m_Good;
-    std::string m_DBName;
-    Options m_Options;
-    leveldb::ExpiryModuleOS * m_Expiry;
-    ExpDB * m_DB;
-    uint64_t m_BaseTime;
-
     void OpenTestDB()
     {
         leveldb::Status status;
@@ -1402,6 +1405,14 @@ public:
         ASSERT_OK(status);
         m_DB->SetClock(m_BaseTime);
     }   // OpenTestDB
+
+protected:
+    bool m_Good;
+    std::string m_DBName;
+    Options m_Options;
+    leveldb::ExpiryModuleOS * m_Expiry;
+    ExpDB * m_DB;
+    uint64_t m_BaseTime;
 
 };  // ExpiryDBTester
 
@@ -1426,7 +1437,7 @@ TEST(ExpiryDBTester, Simple)
     Status s;
     sExpiryDBObject * cursor;
     std::string buffer;
-    leveldb::Iterator * iterator;
+    std::auto_ptr<leveldb::Iterator> iterator;
 
     // enable compaction expiry
     m_Expiry->expiry_enabled=true;
@@ -1450,14 +1461,12 @@ TEST(ExpiryDBTester, Simple)
     }   // for
 
     // verify we can walk it
-    iterator=m_DB->NewIterator(ReadOptions());
+    iterator.reset(m_DB->NewIterator(ReadOptions()));
     for (loop=0, iterator->SeekToFirst(); loop<obj_count; ++loop, iterator->Next())
     {
         ASSERT_EQ(iterator->Valid(), true);
     }   // for
     ASSERT_EQ(iterator->Valid(), false);
-    delete iterator;
-    iterator=NULL;
 
     // expiry set to 2 min, so shift 10
     m_DB->ShiftClockMinutes(10);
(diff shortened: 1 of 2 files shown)


erikleitch:
These changes look ok to me (I have not previously reviewed all of the expiry code, and do have profess to have gone through all of it here).  

Note that class ExpiryDBTester appears to leak m_Expiry in the destructor.  Probably not important in a unit testing context, unless it is ever used in a large loop!

matthewvon:
Erik's observation is quite good.  However there is an underlying "trick" in play that is not obvious (and know that I just now verified via valgrind that the next statement is true).  The "local" pointer to the expiry module, m_Expiry, does not own the object.  The Options object owns the expiry object.  It uses a smart pointer to automatically delete the object once the database closes.

But seriously, good eyes on that one.  I will add a comment to the code in explanation for the future.

matthewvon:
mmaszewski@puma:~/leveldb$ valgrind --leak-check=summary --show-reachable=yes ./expiry_os_test
==31535== Memcheck, a memory error detector
==31535== Copyright (C) 2002-2011, and GNU GPL'd, by Julian Seward et al.
==31535== Using Valgrind-3.7.0 and LibVEX; rerun with -h for copyright info
==31535== Command: ./expiry_os_test
==31535== 
==== Test ExpiryTester.Defaults
==== Test ExpiryTester.MemTableInserterCallback
==== Test ExpiryTester.MemTableCallback
==== Test ExpiryTester.CompactionFinalizeCallback1
==== Test ExpiryTester.OverlapTests
==== Test ExpiryManifestTester.Manifest1
==== Test ExpiryManifestTester.Overlap1
==== Test ExpiryManifestTester.Overlap2
==== Test ExpiryManifestTester.Compact1
==== Test ExpiryDBTester.Simple
==== PASSED 10 tests
==31535== 
==31535== HEAP SUMMARY:
==31535==     in use at exit: 0 bytes in 0 blocks
==31535==   total heap usage: 8,238 allocs, 8,238 frees, 9,884,562 bytes allocated
==31535== 
==31535== All heap blocks were freed -- no leaks are possible
==31535== 
==31535== For counts of detected and suppressed errors, rerun with: -v
==31535== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 4 from 4)

leveldb_os/expiry_os_test.cc
```diff
@@ -1252,7 +1253,7 @@ TEST(ExpiryManifestTester, Overlap2)
 
     // let multiple threads complete
     /// sleep(1) required for Smart OS 1.8 buildbot
-    sleep(1);
+    sleep(2);
```
paulplace, line 1262: What is the significance of changing the sleep() parameter from 1 to 2? Was this just a little extra slop to accommodate the buildbot? Regardless, the comment should be updated to match the code (or vice versa).

leveldb_os/expiry_os_test.cc
```diff
@@ -1349,5 +1350,170 @@ TEST(ExpiryManifestTester, Compact1)
...
+    };
+
+    ~ExpiryDBTester()
+    {
+        // clean up
+        delete m_DB;
+        leveldb::DestroyDB(m_DBName, m_Options);
+    };
+
+    bool m_Good;
```
paulplace: Can add "protected:" above this line. While not necessary since this is a test class, it is still a goodness of OOP.

leveldb_os/expiry_os_test.cc
```diff
@@ -1349,5 +1350,170 @@ TEST(ExpiryManifestTester, Compact1)
...
+          m_BaseTime(port::TimeUint64())
+    {
+        m_DBName = test::TmpDir() + "/expiry";
+
+        // clean up previous execution
+        leveldb::DestroyDB(m_DBName, m_Options);
+
+        m_Options.create_if_missing=true;
+        m_Options.error_if_exists=false;
+        m_Expiry=new leveldb::ExpiryModuleOS;
```
paulplace, line 1385: Per other discussions about the m_Expiry member, it would be good to add a comment that this data member is owned by the m_Options object (so no need to free in this class). That is, m_Expiry is simply a convenience pointer kept around to allow easier updating of the expiry policy in the test below. (This also applies to the m_Expiry data member in the ExpiryManifestTester class above.)

leveldb_os/expiry_os_test.cc
```diff
@@ -1349,5 +1350,170 @@ TEST(ExpiryManifestTester, Compact1)
...
+ * Do simple writes, see if data disappears
+ *
+ */
+TEST(ExpiryDBTester, Simple)
+{
+    size_t obj_count, loop;
+    Status s;
+    sExpiryDBObject * cursor;
+    std::string buffer;
+    leveldb::Iterator * iterator;
```
paulplace: Could change the iterator variable to be a std::auto_ptrleveldb::Iterator. Doing so is far from necessary, since this is test code, but I worry about copying/pasting introducing this pattern into the production code. Since auto_ptr is not really used in the production code either, perhaps this is not a concern.

leveldb_os/expiry_os_test.cc
```diff
@@ -1349,5 +1350,170 @@ TEST(ExpiryManifestTester, Compact1)
...
+    iterator=NULL;
+
+    // expiry set to 2 min, so shift 10
+    m_DB->ShiftClockMinutes(10);
+
+    // all data gone?
+    for (loop=0, cursor=SimpleData; loop<obj_count; ++loop, ++cursor)
+    {
+        s=m_DB->Get(ReadOptions(), cursor->m_Key, &buffer);
+        ASSERT_EQ(s.ok(), false);
```
paulplace: A better test would be ASSERT_EQ(s.IsNotFound(), true);

leveldb_os/expiry_os_test.cc
```diff
@@ -1349,5 +1350,170 @@ TEST(ExpiryManifestTester, Compact1)
...
```
paulplace: A better test would be ASSERT_EQ(s.IsNotFound(), true);

leveldb_os/expiry_os_test.cc
```diff
@@ -1349,5 +1350,170 @@ TEST(ExpiryManifestTester, Compact1)
...
+    }   // for
+
+    iterator=m_DB->NewIterator(ReadOptions());
+    iterator->SeekToFirst();
+    ASSERT_EQ(iterator->Valid(), false);
+    delete iterator;
+    iterator=NULL;
+
```
paulplace, line 1537: With the clock shifted forward 10 minutes, what happens if you compact again? Will it delete the records? Would be interesting to compact, shift the clock back to "now" and see if the records are gone.

leveldb_os/expiry_os_test.cc
matthewvon, line 1262: done

leveldb_os/expiry_os_test.cc
matthewvon, line 1385: argh!  done already but not pushed to github.

leveldb_os/expiry_os_test.cc
matthewvon: done.

leveldb_os/expiry_os_test.cc
matthewvon, line 1537: Test conditions to match desire now at end of function.

paulplace:
+1 eb43eb5

leveldb_os/expiry_os_test.cc
matthewvon: wiki comments updated to include the above additional test of physical delete.

PR #204, opened 2016-08-20
Mv mem fences

matthewvon:
This is one of two parallel branches named mv-mem-fences.  The second branch is within the eleveldb repository.  The two branches must be used together.  

The branches address an extremely rare race condition that causes a protective assert() to fire within eleveldb's refobjects.cc.

A detailed description of the bug and fix is found here:

https://github.com/basho/leveldb/wiki/mv-mem-fences

### DIFF
--- a/db/db_iter.cc
+++ b/db/db_iter.cc
@@ -82,7 +82,9 @@ class DBIter: public Iterator {
     if (kForward==direction_)
     {
       ParsedInternalKey parsed;
-      parsed.type=kTypeValue; parsed.sequence=0; parsed.expiry=0; // clear warnings
+      // this initialization clears a warning.  ParsedInternalKey says
+      //  it is not initializing for performance reasons ... oh well
+      parsed.type=kTypeValue; parsed.sequence=0; parsed.expiry=0;
       ParseInternalKey(iter_->key(), &parsed);
       keymetadata_.m_Type=parsed.type;
       keymetadata_.m_Sequence=parsed.sequence;
--- a/db/db_iter.cc
+++ b/db/db_iter.cc
@@ -82,6 +82,7 @@ class DBIter: public Iterator {
     if (kForward==direction_)
     {
       ParsedInternalKey parsed;
+      parsed.type=kTypeValue; parsed.sequence=0; parsed.expiry=0; // clear warnings
       ParseInternalKey(iter_->key(), &parsed);
       keymetadata_.m_Type=parsed.type;
       keymetadata_.m_Sequence=parsed.sequence;
--- a/util/flexcache.cc
+++ b/util/flexcache.cc
@@ -45,7 +45,8 @@ FlexCache::FlexCache()
     // initialize total memory available based upon system data
     ret_val=getrlimit(RLIMIT_DATA, &limit);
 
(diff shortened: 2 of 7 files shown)


db/db_iter.cc
```diff
@@ -82,6 +82,7 @@ class DBIter: public Iterator {
     if (kForward==direction_)
     {
       ParsedInternalKey parsed;
+      parsed.type=kTypeValue; parsed.sequence=0; parsed.expiry=0; // clear warnings
```
paulplace: The "clear warnings" comment doesn't tell me enough. Is this to avoid compiler warnings about using uninitialized data members? If so, it would be cleaner to initialize the value-type data members in the ParsedInternalKey constructor, which is already defined in dbformat.h.

util/flexcache.cc
```diff
@@ -45,7 +45,8 @@ FlexCache::FlexCache()
     // initialize total memory available based upon system data
     ret_val=getrlimit(RLIMIT_DATA, &limit);
 
-    if (0==ret_val && RLIM_INFINITY!=limit.rlim_max)
+    //  unsigned long caste to fix warning in smartos1.8, smartos 13.1, solaris10
+    if (0==ret_val && (unsigned long)RLIM_INFINITY!=limit.rlim_max)
```
paulplace, line 49: The problem with unconditionally casting RLIM_INFINITY to (unsigned long) is that the types of limit.rlim_max and (unsigned long) might be different sized integers. If one is 32-bit and the other 64-bit on some platform, then I'm not sure you'll get the comparison you want, since unsigned values typically get converted to signed integers during promotion (just depends on the compiler); then you might not get the comparison you intend due to sign extension.

Or this might "just work" on all the platforms we currently support.

If the only platforms complaining about this are the Solaris based compilers (as the comment indicates), could do the cast inside an "#ifdef SOLARIS" conditional block.

util/refobject_base.h
```diff
@@ -51,16 +51,17 @@ namespace leveldb {
...
 
  public:
     RefObjectBase() : m_RefCount(0) {}
     virtual ~RefObjectBase() {}
 
-    uint32_t RefInc() {return(inc_and_fetch(&m_RefCount));}
+    virtual uint32_t RefInc() {return(inc_and_fetch(&m_RefCount));}
 
-    uint32_t RefDec()
+    virtual uint32_t RefDec()
```
paulplace, line 64: Since ErlRefObject overrides RefDec(), I wonder if not having RefObjectBase::RefDec() be virtual could have been a problem prior to this change?

db/db_iter.cc
matthewvon: ... but the comment on the constructor says that they are not initializing there for performance reasons.  Upgrading comment in db_iter.cc.

util/refobject_base.h
matthewvon, line 64: I reviewed the code history.  eleveldb's RefObject used to be the base class for ErlRefObject.  It did have RefDec() as a virtual.  When the base class was unified with the "equivalent" in leveldb, no one noticed that leveldb's RefDec() was not virtual.

So yes, maybe there were side effects.  If so, they are now fixed ...

paulplace:
+1 2f0b293

PR #205, opened 2016-09-13
Only switch manifest format if expiry actually used

matthewvon:
Actively set second param of VersionEdit.EncodeTo() based upon whether or not expiry module is active.  The flag controls whether manifest is traditional format or new Expiry enabled format which is NOT backward compatible.

Implementation created a generic ExpiryActivated() const function in the Options and ExpiryModule classes.  Adjusted code everywhere to use this accessor function instead of explicit tests.


fadushin:
+1 9d260b8

PR #206, opened 2016-09-14
Added unit test support for validating manifest file format 1 versus 2.

matthewvon:
Decided the implicit tests of manifest format 1 versus 2 were insufficient.  Created an explicit method for validating VersionEdit::EncodeTo()'s second parameter.  Then updated existing VersionEdit unit test to use the validation.


paulplace:
+1 dfa6331

PR #207, opened 2016-09-15
Do not read bloom filter upon .sst open

matthewvon:
Bloom filters can approach 600Kbytes.  They are not used during compaction or iterator only work.  A Get() request that opens a .sst file would likely process fewer bytes by actually reading the potential data block, than first unpacking the bloom filter.  Memory tight machines might never open the bloom filter if jumping rapidly between .sst table files.

This branch reads the bloom filter only upon the second read of a .sst table file.

This branch also corrects a write throttle problem detected during testing of the bloom filter change.

Full details here:  
  https://github.com/basho/leveldb/wiki/mv-delayed-bloom

### DIFF
--- a/db/table_cache.cc
+++ b/db/table_cache.cc
@@ -100,18 +100,24 @@ Status TableCache::FindTable(uint64_t file_number, uint64_t file_size, int level
       if (!for_iterator && table->ReadFilter())
       {
           // TableAndFile now going to be present in two cache entries
+	  //  1. retrieve old entry within file cache
           TableAndFile* tf = reinterpret_cast<TableAndFile*>(cache_->Value(*handle));
           inc_and_fetch(&tf->user_count);
 
-          //  must clean file size, do not want double count
+          //  2. must clean file size, do not want double count
           if (NULL!=tf->doublecache)
               tf->doublecache->SubFileSize(tf->table->GetFileSize());
 
+	  //  3. release current reference (and possible special overlap reference)
           cache_->Release(*handle);
           if (tf->level<config::kNumOverlapLevels)
               cache_->Release(*handle);
 
+	  //  4. create second table cache entry using TableObjectSize that now includes
+	  //     bloom filter size
           *handle = cache_->Insert(key, tf, table->TableObjectSize(), &DeleteEntry);
+
+	  //  5. set double reference if an overlapped file (prevents from being flushed)
           if (level<config::kNumOverlapLevels)
               cache_->Addref(*handle);
 
--- a/db/table_cache.cc
+++ b/db/table_cache.cc
@@ -19,11 +19,12 @@ static void DeleteEntry(const Slice& key, void* value) {
   TableAndFile* tf = reinterpret_cast<TableAndFile*>(value);
   if (0==dec_and_fetch(&tf->user_count))
   {
-      if (NULL!=tf->doublecache)
-          tf->doublecache->SubFileSize(tf->table->GetFileSize());
-      delete tf->table;
-      delete tf->file;
-      delete tf;
+    if (NULL!=tf->doublecache)
+      tf->doublecache->SubFileSize(tf->table->GetFileSize());
+    delete tf->table;
+    delete tf->file;
+    delete tf;
+    gPerfCounters->Inc(ePerfDebug1);
   }   // if
 }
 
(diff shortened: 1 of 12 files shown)


db/table_cache.cc
```diff
@@ -90,6 +94,36 @@ Status TableCache::FindTable(uint64_t file_number, uint64_t file_size, int level
   }
   else
   {
+      Table *table = reinterpret_cast<TableAndFile*>(cache_->Value(*handle))->table;
+
+      // this is NOT first access, see if bloom filter can load now
+      if (!for_iterator && table->ReadFilter())
+      {
+          // TableAndFile now going to be present in two cache entries
+	  //  1. retrieve old entry within file cache
```
paulplace: The new code in this file uses 4-space tab settings, whereas the original code uses 2-space tabs. This is probably okay, but more problematic is the mix of spaces and hard-tabs (on several of the numbered comments in this section), which causes alignment ugliness in viewers with different tab stop settings.

tools/sst_scan.cc
```diff
@@ -135,6 +135,7 @@ main(
                     tot_size=0;
 
                     table = reinterpret_cast<leveldb::TableAndFile*>(table_cache->TEST_GetInternalCache()->Value(fhandle))->table;
+		    table->ReadFilter();
```
paulplace: Have hard-tabs at the beginning of this line, whereas the rest of the file uses spaces; this causes alignment ugliness in my viewer.

table/table.cc
```diff
@@ -143,8 +149,12 @@ void Table::ReadMeta(const Footer& footer) {
               iter->Seek(key);
               if (iter->Valid() && iter->key() == Slice(key))
               {
-                  ReadFilter(iter->value(), policy);
-                  gPerfCounters->Inc(ePerfBlockFilterRead);
+		  // store information needed to load bloom filter
```
paulplace: Lines 152-154 have hard-tabs instead of spaces.

db/version_set.cc
```diff
@@ -1307,14 +1307,15 @@ VersionSet::UpdatePenalty(
...
-            if (config::kL0_CompactionTrigger < v->files_[level].size())
+            if (config::kL0_SlowdownWritesTrigger < v->files_[level].size())
             {
                 // assume each overlapped file represents another pass at same key
                 //   and we are "close" on compaction backlog
                 if ( v->files_[level].size() < config::kL0_SlowdownWritesTrigger)
                 {
-                    value = (v->files_[level].size() - config::kL0_CompactionTrigger);
-                    count=0;
+                    // this code block will not execute due both "if"s using same values now
```
paulplace, line 1316: Since this block will not execute, shouldn't it be removed?

db/version_set.cc
```diff
@@ -1358,9 +1359,11 @@ VersionSet::UpdatePenalty(
             {   // light penalty
                 count=(int)(static_cast<double>(level_bytes) / gLevelTraits[level].m_DesiredBytesForLevel);
                 value=count; // this approximates the number of compactions needed, no other penalty
+                value/=10;  // allow 10 files, approx 2G without throttle
```
paulplace, line 1362: The comment seems to indicate this should be something like `if (value>10) value -= 10; else value = 0;` whereas the current code does a 10-fold reduction.

db/table_cache.cc
```diff
@@ -90,6 +94,36 @@ Status TableCache::FindTable(uint64_t file_number, uint64_t file_size, int level
...
+
+	  //  4. create second table cache entry using TableObjectSize that now includes
+	  //     bloom filter size
+          *handle = cache_->Insert(key, tf, table->TableObjectSize(), &DeleteEntry);
+
+	  //  5. set double reference if an overlapped file (prevents from being flushed)
+          if (level<config::kNumOverlapLevels)
+              cache_->Addref(*handle);
+
+          gPerfCounters->Inc(ePerfDebug0);
```
paulplace: I only see this counter being incremented; it would be nice to have a perf counter that tracks the total number of these entries so that you can ensure the reference counting logic is correct (would need to decrement such a counter in DeleteEntry(), I think).

paulplace (commented):
Finished code review

db/table_cache.cc
matthewvon: fixed ... laptop must lack proper .emacs settings.  checked other three machines used during testing.

db/table_cache.cc
matthewvon: I will put one of the other debug counters within delete entry for retest.  SNMP only knows these as counters, not a gauge.  So will have to compare the two.

db/version_set.cc
matthewvon: I may change my mind and activate it.  The whole throttle may need another tuning based upon bugs found in Riak today (if the bugs are fixed).

db/version_set.cc
matthewvon: Current goal is one "value" for each 10 files of backlog.  Number of files is also an approximation based upon size.  Purely a trial and error value.

table/table.cc
matthewvon: corrected

tools/sst_scan.cc
matthewvon: corrected.  thank you.

paulplace:
+1 84684e8

PR #212, opened 2016-10-18
Dynamic decision to swap threads or not on Erlang Write operation

matthewvon:
eleveldb swaps each Erlang request from an Erlang scheduler thread to a local eleveldb worker thread. This is to prevent Erlang scheduler collapses. Write operations do not actually need to swap all the time. This branch creates the capability of dynamic selection to swap threads or not.

Detailed discussion is here:

https://github.com/basho/leveldb/wiki/mv-nonblocking-write2

### DIFF
--- a/db/db_impl.cc
+++ b/db/db_impl.cc
@@ -189,7 +189,8 @@ DBImpl::DBImpl(const Options& options, const std::string& dbname)
       level0_good(true),
       throttle_end(0),
       running_compactions_(0),
-      block_size_changed_(0), last_low_mem_(0)
+      block_size_changed_(0), last_low_mem_(0),
+      non_block_tickets_(0), last_penalty_(0)
 {
   current_block_size_=options_.block_size;
   DBList()->AddDB(this, options_.is_internal_db);
@@ -1755,7 +1756,8 @@ Status DBImpl::Write(const WriteOptions& options, WriteBatch* my_batch) {
   }
 
   // May temporarily unlock and wait.
-  status = MakeRoomForWrite(my_batch == NULL);
+  if (!options.non_blocking)
+      status = MakeRoomForWrite(my_batch == NULL);
   uint64_t last_sequence = versions_->LastSequence();
   Writer* last_writer = &w;
   if (status.ok() && my_batch != NULL) {  // NULL batch is for compactions
@@ -1802,7 +1804,15 @@ Status DBImpl::Write(const WriteOptions& options, WriteBatch* my_batch) {
   gPerfCounters->Inc(ePerfApiWrite);
 
   // protect use of versions_ ... still within scope of mutex_ lock
-  throttle=versions_->WriteThrottleUsec(IsCompactionScheduled());
+  if (0==add_and_fetch(&non_block_tickets_, (uint32_t)0))
+  {
+      throttle=versions_->WriteThrottleUsec();
+  }   // if
+  else
+  {
+      dec_and_fetch(&non_block_tickets_);
+      throttle=0;
+  }   // else
   }  // release  MutexLock l(&mutex_)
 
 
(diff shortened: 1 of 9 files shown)


db/db_impl.cc
```diff
@@ -1965,6 +1983,34 @@ WriteBatch* DBImpl::BuildBatchGroup(Writer** last_writer) {
...
+ */
+bool
+DBImpl::RequestNonBlockTicket()
+{
+    bool ret_flag(false);
+
+    // assume ticket succeeds
+    inc_and_fetch(&non_block_tickets_);
+
+    // est_mem_usage_ is last known mem_->ApproximateMemoryUsage(), write_buffer_size is constant
```
paulplace, line 2000: This is purely a matter of style, but I would find this a bit easier to read if it were a single conditional; something like

``` C++
if ( est_mem_usage_ <= options_.write_buffer_size
     && 1==GetThrottleWriteRate()
     && 0==GetLastPenalty() )
   ret_flag = true;
else
   dec_and_fetch(&non_block_tickets_);
```

db/db_impl.h
```diff
@@ -88,7 +89,12 @@ class DBImpl : public DB {
   bool IsCompactionScheduled();
   uint32_t RunningCompactionCount() {mutex_.AssertHeld(); return(running_compactions_);};
 
- protected:
+  // memory barrier set/retrieval of last_penalty_
+  uint64_t GetLastPenalty() {return(add_and_fetch(&last_penalty_, (uint64_t)0));};
+  void SetLastPenalty(uint64_t NewPenalty)
+  {compare_and_swap(&last_penalty_, (uint64_t)last_penalty_, NewPenalty);};
```
paulplace, line 95: hmm...not sure if it's an issue that we're reading the value of last_penalty_ without a memory barrier (for the second parameter to compare_and_swap())

db/db_impl.h
```diff
@@ -223,13 +229,15 @@ class DBImpl : public DB {
   volatile size_t current_block_size_;    // last dynamic block size computed
   volatile uint64_t block_size_changed_;  // NowMicros() when block size computed
   volatile uint64_t last_low_mem_;        // NowMicros() when low memory last seen
+  volatile bool hotbackup_pending_;       // true if hotbackup cycle initiated, blocks close
```
paulplace, line 232: I recommend moving this bool variable to the end of the list of new variables, and moving the two 64-bit variables up before non_blocking_tickets_ (which is 32-bits). Doing so will reduce wasted space when the compiler adds packing for these variables.

If you really want to get ambitious, you could reorder this entire block of member variables so that the 64-bit variables are together, followed by the 32-bit variables, etc.

None of this is a big deal, since we won't have many DBImpl objects allocated at one time; it's just a general habit I've adopted over the years.

db/db_impl.h
matthewvon: Reasonable question.  I think I will drop "GetLastPenalty()" into that second spot.

include/leveldb/db.h
```diff
@@ -171,6 +171,12 @@ class DB {
   // eleveldb, access to leveldb's logging routines.
   virtual Logger* GetLogger() const { return NULL; }
 
+  // Riak specific function:  Request a "ticket" that allows
+  // bypass of most wait / blocking activity in a Write() call.
+  // Used by eleveldb to potential skip a thread switch prior
```
paulplace, line 176: "to potentially"

db/db_impl.h
matthewvon: I'll shift stuff as long as it does not cause grief in the constructor (which it should not).

util/throttle.cc
```diff
@@ -235,15 +235,20 @@ ThrottleThread(
             }   // else
 
             // change the throttle slowly
+            temp_throttle=gThrottleRate;
```
paulplace, line 238: If you're concerned about accessing gThrottleRate with a memory barrier, it seems like you should use GetThrottleWriteRate() instead of the raw variable (in all the places it shows up below). Either that, or call GetThrottleWriteRate() once and then cache that value locally and use the cached value as you adjust temp_throttle.

db/db_impl.h
paulplace: The only issue to consider with respect to the constructor is to ensure that the data member initialization list in the ctor initializes the members in the same order that they're declared. If you don't, some compilers will issue a warning.

db/db_impl.cc
```diff
@@ -1965,6 +1983,34 @@ WriteBatch* DBImpl::BuildBatchGroup(Writer** last_writer) {
...
+    ret_flag = (est_mem_usage_ <= options_.write_buffer_size);
```
paulplace, line 2001: The wiki entry for this change mentions testing that the write buffer is not "too close to full". Here we're only checking if we've hit the limit. Should this comparison be adjusted to allow for a little extra space?

db/db_impl.cc
```diff
@@ -1806,7 +1807,9 @@ Status DBImpl::Write(const WriteOptions& options, WriteBatch* my_batch) {
   }
 
   // May temporarily unlock and wait.
-  status = MakeRoomForWrite(my_batch == NULL);
+  if (0==add_and_fetch(&non_block_tickets_, (uint32_t)0))
+      status = MakeRoomForWrite(my_batch == NULL);
```
paulplace, line 1811: What happens if non_block_tickets_ is non-0, so that we don't call MakeRoomForWrite(), but there's not enough space for all items in the write batch?

That is, is it strictly necessary to call MakeRoomForWrite() before writing the next record, or will the memory footprint just grow beyond the desired limit, at which point blocking writes will be required?

matthewvon: write_buffer_size is a threshold, not a limit.  ApproximateMemoryUsage() is not an exact estimate of how much space a memory buffer (skip list) will require on disk (before compression).  Essentially, this is simply the same rule used in MakeRoomForWrite() to initiate a purge of the current write buffer.  I considered sending in the size of this write and tracking "upcoming writes" size.  That did not seem to provide benefit since the logic blocks a purge until all the known writes complete.  Again "threshold" not "limit".

paulplace: Makes sense. I just wanted to be sure.

matthewvon:
branch mv-nonblocking-write3 rebuilt for merge.

PR #213, opened 2016-10-26
Exclude Riak's "metadata" key from global expiry

matthewvon:
Riak creates a special metadata key within its databases (vnodes).  This key needs to be explicitly excluded from global expiry.

Detail documentation is here:
    https://github.com/basho/leveldb/wiki/mv-no-md-expiry


fadushin:
Reviewed in person with MVM

+1

PR #214, opened 2016-10-29
Adjust rules for non-penalty compaction backlog

matthewvon:


fadushin:
MVM to add defensive asserts to avoid division by 0 (which is not technically possible now, given the declaration of static tables, but does give us some future proofing.

PR #216, opened 2016-11-18
Mv bucket expiry

matthewvon:
... more to come.

### DIFF
--- a/db/penalty_test.cc
+++ b/db/penalty_test.cc
@@ -39,7 +39,7 @@ int main(int argc, char** argv)
 
 namespace leveldb {
 
-class TestVersion : public Version    
+class TestVersion : public Version
 {
 public:
     TestVersion()
@@ -53,7 +53,7 @@ public:
             m_LevelFileCount[loop]=0;
         }   // for
     };
-    
+
     virtual size_t NumFiles(int level) const {return(m_LevelFileCount[level]);};
 
     virtual const std::vector<FileMetaData*> & GetFileList(int level) const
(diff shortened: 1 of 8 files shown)


db/penalty_test.cc
```diff
@@ -108,11 +108,20 @@ TEST(PenaltyTester, Debug1)
     
     UpdatePenalty(&version);
 
-    ASSERT_EQ(version.WritePenalty(), 0);
+    ASSERT_TRUE(0==version.WritePenalty());
 
 }   // test Debug1
 #endif
 
+  /*******************
```
paulplace: I recommend moving this comment block above the first test (Debug1) since it applies to that test, too (even though it is currently commented out).

tools/sst_scan.cc
```diff
@@ -241,13 +267,36 @@ main(
                                     if (riak_translations && '\x10'==*parsed.user_key.data())
                                     {
                                         leveldb::Slice cursor_slice;
+                                        std::string type, bucket;
```
paulplace: These variables are not currently used, due to the "#if 0" blocks below; should probably comment out this line for now.

util/expiry_os.h
```diff
@@ -40,7 +40,6 @@ class ExpiryModuleOS : public ExpiryModule
     ExpiryModuleOS()
```
paulplace, line 40: Comments in your wiki indicate that you already know this, but just in case: you can make this ctor (and that of ExpiryModuleEE) protected and declare ExpiryModule::CreateExpiryModule() as a friend, which forces everyone to allocate via your factory method. The only complication is in the unit tests, where you would need to update them to allocate via the factory method.

db/penalty_test.cc
matthewvon: done

tools/sst_scan.cc
matthewvon: moved inside #if 0/#endif block

util/expiry_os.h
matthewvon: agreed ... but saving the unit test rework for next branch (which likely starts later today).

paulplace:
+1 51f0c7b

PR #222, opened 2017-02-13
Address case where non-tiered storage hot backup will not delete backup.4 directory.

matthewvon:
The hot backup code was extensively tested with the more difficult configuration of tiered storage.  Therefore a bug in the simpler non-tiered storage was missed.  The routine MakeTieredDbname() in db/filename.cc was originally designed to populate fast and slow tier names of the Options structure given a user's database path.

Hot backup is an internal user.  It does not know how tiered storage was or was not constructed from components.  Therefore, hot backup calls DestroyDB() (db/db_impl.cc) without a database name but with a full populated Options.  This works just fine for tiered storage.  It failed with non-tiered because MakeTieredDbname() returned a zero length string ... the string was therefore useless for deleting the backup.4 directory.

The PR adds a third case to MakeTieredDbname() and creates a unit test for the case within db/filename_test.cc


hazen:
@matthewvon It looks like you need to rebase to resolve your conflicts with `develop`

matthewvon:
@javajolt I will fix the rebase issues upon +1 ... honestly unclear what got screwed up.  Just need the code approved.

matthewvon:
@javajolt & @lukebakken This looks like the "develop" branch was instead git hub tag of riak-2.2.0's release.  I will have to move db/filename.cc and db/filename_test.cc to a new branch.

If reviewing please focus on those two files.

matthewvon:
this is likely a process error on my part ... but git still sucks.

PR #223, opened 2017-02-14
Address case where non-tiered storage hot backup will not delete backup.4 directory.

matthewvon:
The hot backup code was extensively tested with the more difficult configuration of tiered storage. Therefore a bug in the simpler non-tiered storage was missed. The routine MakeTieredDbname() in db/filename.cc was originally designed to populate fast and slow tier names of the Options structure given a user's database path.

Hot backup is an internal user. It does not know how tiered storage was or was not constructed from components. Therefore, hot backup calls DestroyDB() (db/db_impl.cc) without a database name but with a full populated Options. This works just fine for tiered storage. It failed with non-tiered because MakeTieredDbname() returned a zero length string ... the string was therefore useless for deleting the backup.4 directory.

The PR adds a third case to MakeTieredDbname() and creates a unit test for the case within db/filename_test.cc

### DIFF
--- a/db/filename.cc
+++ b/db/filename.cc
@@ -282,7 +282,7 @@ MakeTieredDbname(
     Options & options)             // input/output ... writable Options, tiered values changed
 {
     // case for "", used with internal calls to DestroyDB
-    if (0==dbname.length() && 0!=options.tiered_fast_prefix.length())
+    if (0==dbname.size() && 0!=options.tiered_fast_prefix.size())
     {
         // do NOTHING ... options already initialized
     }   // if
--- a/db/filename.cc
+++ b/db/filename.cc
@@ -281,7 +281,12 @@ MakeTieredDbname(
     const std::string & dbname,    // input ... original dbname from DBImpl constructor
     Options & options)             // input/output ... writable Options, tiered values changed
 {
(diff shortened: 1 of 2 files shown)


erikleitch:
This is a simple fix to add a check for empty dbname in MakeTieredDbname() when options.tiered_fast_prefix.length() > 0.

The change itself seems fine to me, provided the underlying assumption is correct: ie, that hot backup always calls MakeTieredDbname with an _empty_ string and _non-empty_ fast_prefix, and that otherwise MakeTieredDbname gets called with a _non-empty_ string and _empty_ fast_prefix.  This appears to be the case in hot_backup_test.cc, but without knowing more about tiered storage and hot backups, I can't verify the validity of that assumption in general.  But I assume @matthewvon knows that that is the case.

I've certainly verified that the tests work as advertised (and that they don't work if you remove the check) 

Totally minor note: I'd avoid mixing std::string::size() and std::string::length() in the same piece of code.  They're synonymous, but may as well pick one and stick with it.

matthewvon:
std::string::size() now used exclusively within MakeTieredDbname().  That is a Basho created routine.  Did not review size() versus length() in remainder of Google created routines.

leveldb_ee tagging rebuilt.  It was messed up when this branch was rebuilt.

hazen:
Unit tests now build and pass locally

hazen:
:+1:

Issue #237, opened 2020-05-13
Add ppc64le support


seth-priya:
The build fails with the following error
 #error Please implement AtomicPointer for this platform.
  ^~~~~

seth-priya:
Confirmed that https://github.com/basho/leveldb/pull/235 fixes this issue, please review

PR #238, opened 2020-06-29
Changes to support ppc64le

seth-priya:
Add support for IBM Power Linux platform, build and all unit tests are passing.


seth-priya:
Reference: https://github.com/basho/leveldb/issues/237

PR #242, opened 2022-05-20
Unbreak 32-bit build

hmmr:
This PR:
* replaces deprecated `auto_ptr` with `unique_ptr`;
* drops some [dead code](73837fe32fa84ddf) (port/port_chromium.h is not present in the project and `LEVELDB_PLATFORM_CHROMIUM` is never defined);
* [enables](1f69b6ec0ed02e3a) building on 32-bit arches, by dropping workarounds in various add_and_fetch and dec_and_fetch and directly falling through to gcc built-ins. Specifically, [x86](https://gitlab.alpinelinux.org/hmmr/aports/-/jobs/724044), where it also passes the tests.

Eleveldb tests are [passing](https://github.com/TI-Tokyo/eleveldb/runs/6474322059?check_suite_focus=true).

### DIFF
--- a/port/port.h
+++ b/port/port.h
@@ -8,13 +8,8 @@
 #include <string.h>
 #include "leveldb/ldb_config.h"
 
-// Include the appropriate platform specific file below.  If you are
-// porting to a new platform, see "port_example.h" for documentation
-// of what the new port_<platform>.h file must provide.
-#if defined(LEVELDB_PLATFORM_POSIX)
-#  include "port/port_posix.h"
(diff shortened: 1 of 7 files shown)


martinsumner:
Was the change to table_test.cc due to an intermittent test failure?

hmmr:
It wasn't even intermittent (it was reliably failing all the times I ran the tests). https://github.com/basho/leveldb/pull/232 suggests it was failing all along. I was unable to dig deeper and find the root cause of it failing; I can only confirm that the upgrade of snappy didn't make it worse.

martinsumner (approved):

